PR #2, opened 2018-12-04
Create project, parse and upload gigaword documents from a single Gzip file

shinyichen:


gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/GigawordParserUtils.java
```diff
@@ -0,0 +1,126 @@
...
+    return fullText.toString();
+  }
+
+  /**
+   * split full text into separate articles
+   *
+   * @param fullText
+   * @return A HashMap of document ID and article
+   */
+  private static HashMap<String, String> splitTextIntoArticles(String fullText) {
```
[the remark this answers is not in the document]
rgabbard: I wouldn't change it now, but an alternate approach to consider is just returning an `Iterator` over the articles in the file, since you don't have any need for the random access nature of the `Map`. That would let you avoid loading the entire files into memory.  (In this case we know the files are small enough we can get away with it)

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/GigawordIndexer.java
```diff
@@ -190,6 +199,11 @@ public static void main(String[] args) {
     builder.properties("running.from.commandline=true");
     builder.bannerMode(Banner.Mode.OFF);
     builder.initializers(new GigawordApplicationContextInitializer());
+
+    // need this line to set user home directory to /.inception
+    // otherwise it will be defaulted to ./webanno and documents will be uploaded here
+    // and Inception app won't be able to find uploaded documents.
+    SettingsUtil.customizeApplication("inception.home", ".inception");
```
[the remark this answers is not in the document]
shinyichen: Where documents are uploaded to will depend on this. Otherwise it defaults to .webanno, and when we run Inception, it will not find the uploaded documents.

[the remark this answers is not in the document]
rgabbard: How does the value of this property translate to a concrete file system path?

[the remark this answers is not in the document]
shinyichen: It updates the string property in the SettingsUtil, which is used by Webanno when setting up repository directory.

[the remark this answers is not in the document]
rgabbard: How does Webanno interpret this value? For example, does it create a sub-directory with that name somewhere? Under the user's home directory or elsewhere?  Can you supply an absolute path? If so, what happens?

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/GigawordParserUtils.java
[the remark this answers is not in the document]
shinyichen: I don't understand what you mean in the first comment, can you explain?

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/GigawordIndexer.java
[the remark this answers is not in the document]
shinyichen: A file directory is constructed using those properties, like below.
```
new File(System.getProperty(SettingsUtil.getPropApplicationHome(),
                System.getProperty("user.home") + "/"
                        + SettingsUtil.getApplicationUserHomeSubdir()),
                "repository");
```
You can see it here.
https://github.com/webanno/webanno/blob/master/webanno-api-dao/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/dao/RepositoryProperties.java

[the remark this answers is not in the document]
rgabbard: Ok, so this will use the path pointed to by `inception.home` if present and will otherwise fall back on `~/.inception`. Got it

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/IndexGigawordWithElasticSearch.java
```diff
@@ -1,52 +1,83 @@
...
+      System.exit(-1);
+    }
+
+    // start indexing
+    try {
+
+      RestHighLevelClient client =
+          new RestHighLevelClient(
+              RestClient.builder(
+                  new HttpHost("localhost", 9200, "http"),
```
rgabbard: Make the ports and host name configurable by parameters, please

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/IndexGigawordWithElasticSearch.java
```diff
@@ -1,52 +1,83 @@
...
+        Article article = iterator.next();
+
+        bulkRequest.add(
+            new IndexRequest(indexName, "doc", article.getId()).source("text", article.getText()));
+      }
+
+      client.bulk(bulkRequest, RequestOptions.DEFAULT);
+
+      client.close();
+      client = null;
```
rgabbard: You shouldn't need to do this - even if anything is to be gained from releasing the reference to `client` early, it will go out of scope at the end of the block anyway.

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/IndexGigawordWithElasticSearch.java
```diff
@@ -1,52 +1,83 @@
...
+
+    } catch (Exception e) {
+      log.error(e.getMessage());
+      e.printStackTrace();
+      System.exit(-1);
```
rgabbard: `System.exit(1)` is more idiomatic. Negative exit codes could cause weird results for some shells.

rgabbard: I think you can just do `log.error("Caught exception: {}", e)`

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/IndexGigawordWithElasticSearch.java
```diff
@@ -1,52 +1,83 @@
...
+      RestHighLevelClient client =
+          new RestHighLevelClient(
+              RestClient.builder(
+                  new HttpHost("localhost", 9200, "http"),
+                  new HttpHost("localhost", 9201, "http")));
+
+      BulkRequest bulkRequest = new BulkRequest();
+      GigawordFileProcessor proc = new GigawordFileProcessor(file);
+      Iterator<Article> iterator = proc.iterator();
+      while (iterator.hasNext()) {
```
rgabbard: We probably don't want send bulk requests of unbounded size.  Use Guava's `Iterators.partition` to send, say, 100 documents at a time: https://google.github.io/guava/releases/15.0/api/docs/com/google/common/collect/Iterators.html#partition(java.util.Iterator,%20int)

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/GigawordFileProcessor.java
```diff
@@ -0,0 +1,102 @@
...
+    private int startNextSearchAt = 0;
+
+    private ArticlesIterator(String text) {
+      fullText = text;
+    }
+
+    @Override
+    protected Article computeNext() {
+
+      if (startNextSearchAt >= fullText.length()) return endOfData();
```
rgabbard: {}s again

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/GigawordFileProcessor.java
```diff
@@ -0,0 +1,102 @@
...
+        startNextSearchAt = endOfDoc + 1;
+
+        // find document id
+        Matcher m =
+            GIGAWORD_DOC_ELEMENT_PATTERN.matcher(
+                docString.substring(0, Math.min(100, docString.length())));
+        if (m.find()) {
+          String docId = m.group(1);
+          return new Article(docId, docString);
+        }
```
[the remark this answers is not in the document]
shinyichen: I changed this to a while loop. If found document with ID, returns the `Article`. Otherwise it will go on go find the next document with ID.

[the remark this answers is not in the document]
rgabbard: This ought not to happen, so I think we would prefer to raise an exception

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/IndexGigawordWithElasticSearch.java
```diff
@@ -1,52 +1,151 @@
...
+
+  private static final String PARAM_FILEPATH = "filePath";
+
+  private static final String PARAM_INDEX_NAME = "indexName";
+
+  private static final String PARAM_HOSTNAME_PRI = "primaryHostName";
+
+  private static final String PARAM_HOSTNAME_SEC = "secondaryHostName";
+
+  private static final String PARAM_PORT_PRI = "primaryPort";
```
rgabbard: Prefer to spell out `PRIMARY` and `SECONDARY`.  Bytes are cheap, confusion is expensive :-)

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/IndexGigawordWithElasticSearch.java
```diff
@@ -1,52 +1,151 @@
...
-                client.index(new IndexRequest("documents", "doc", docId)
-                        .source("text", docContent));
-            }
-
-            // TODO: in practice, you will want to use BulkRequest for speed
-        }
+
+  private static final Logger log = LoggerFactory.getLogger(IndexGigawordWithElasticSearch.class);
+
+  private static final String DEFAULT_INDEX_NAME = "Gigaword";
```
rgabbard: To avoid errors, don't provide a default for the index name

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/IndexGigawordWithElasticSearch.java
```diff
@@ -1,52 +1,151 @@
...
+
+  private static final String DEFAULT_HOST = "localhost";
+
+  private static final int DEFAULT_PORT_PRI = 9200;
+
+  private static final int DEFAULT_PORT_SEC = 9201;
+
+  private static final String PARAM_FILEPATH = "filePath";
```
rgabbard: Perhaps call this `gigawordFilePath` for greater clarity

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/IndexGigawordWithElasticSearch.java
```diff
@@ -1,52 +1,151 @@
...
+  private static final String PARAM_HOSTNAME_PRI = "primaryHostName";
+
+  private static final String PARAM_HOSTNAME_SEC = "secondaryHostName";
+
+  private static final String PARAM_PORT_PRI = "primaryPort";
+
+  private static final String PARAM_PORT_SEC = "secondaryPort";
+
+  /**
+   * @param argv takes two arguments, a path to the GZip file and an index name
```
rgabbard: Comment is out-dated

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/IndexGigawordWithElasticSearch.java
```diff
@@ -1,52 +1,151 @@
...
+  public static void main(String[] argv) throws IOException {
+
+    // get parameter file
+    final Parameters parameters;
+
+    if (argv.length > 0) {
+      parameters = Parameters.loadSerifStyle(new File(argv[0]));
+    } else {
+      log.error("Input file required");
+      return;
```
rgabbard: `System.exit(1)`

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/IndexGigawordWithElasticSearch.java
```diff
@@ -1,52 +1,151 @@
...
+   */
+  public static void main(String[] argv) throws IOException {
+
+    // get parameter file
+    final Parameters parameters;
+
+    if (argv.length > 0) {
+      parameters = Parameters.loadSerifStyle(new File(argv[0]));
+    } else {
+      log.error("Input file required");
```
rgabbard: Also, a clearer error message would be something like `"Expected one argument, a parameter file."`

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/IndexGigawordWithElasticSearch.java
```diff
@@ -1,52 +1,151 @@
...
+      return;
+    }
+
+    // validate and parse inputs
+    parameters.assertExactlyOneDefined(PARAM_FILEPATH);
```
rgabbard: I don't understand this?

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/IndexGigawordWithElasticSearch.java
```diff
@@ -1,52 +1,151 @@
...
+
+    String indexName = parameters.getOptionalString(PARAM_INDEX_NAME).or(DEFAULT_INDEX_NAME);
+    File file = parameters.getExistingFile(PARAM_FILEPATH);
```
rgabbard: Use a more informative name (e.g. `concatenatedFileToIndex`)

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/IndexGigawordWithElasticSearch.java
```diff
@@ -1,52 +1,151 @@
...
+
+    String contentType = Files.probeContentType(file.toPath());
+    if ((contentType != null && !contentType.equalsIgnoreCase("application/gzip"))
+        || (contentType == null
+            && !FilenameUtils.getExtension(file.getName()).equalsIgnoreCase("gz"))) {
+      log.error("{} is not a valid GZip file", file.getAbsolutePath());
+      return;
```
rgabbard: `System.exit(1)`. Return will produce a successful exit status.

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/IndexGigawordWithElasticSearch.java
```diff
@@ -1,52 +1,151 @@
...
+                      parameters.getOptionalPositiveInteger(PARAM_PORT_SEC).or(DEFAULT_PORT_SEC),
+                      "http")));
+
+      // indexing
+      index(client, file, indexName);
+      client.close();
+
+    } catch (Exception e) {
+      log.error("Caught exception: {}", e);
+      return;
```
rgabbard: exit status issue again

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/IndexGigawordWithElasticSearch.java
```diff
@@ -1,52 +1,151 @@
...
+    }
+  }
+
+  public static void index(RestHighLevelClient client, File file, String indexName)
+      throws IOException {
+    BulkRequest bulkRequest = new BulkRequest();
+    GigawordFileProcessor proc = new GigawordFileProcessor(file);
+    Iterator<Article> iterator = proc.iterator();
+
+    // divide iterator to send a bounded number of bulk requests.
```
rgabbard: The number of bulk requests is unbounded, but they are of bounded size

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/IndexGigawordWithElasticSearch.java
```diff
@@ -1,52 +1,151 @@
...
+
+    } catch (Exception e) {
+      log.error("Caught exception: {}", e);
+      return;
+    }
+  }
+
+  public static void index(RestHighLevelClient client, File file, String indexName)
+      throws IOException {
+    BulkRequest bulkRequest = new BulkRequest();
```
rgabbard: This `BulkRequest` needs to be constructed inside the while loop - otherwise, it just keeps getting bigger and bigger each time.

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/IndexGigawordWithElasticSearch.java
```diff
@@ -1,52 +1,151 @@
...
+    while (partitions.hasNext()) {
+      List<Article> articles = partitions.next();
+      for (Article article : articles) {
+        bulkRequest.add(
+            new IndexRequest(indexName, "doc", article.getId()).source("text", article.getText()));
+      }
+
+      BulkResponse bulkResponse = client.bulk(bulkRequest, RequestOptions.DEFAULT);
+      if (bulkResponse.hasFailures()) {
+        log.error(bulkResponse.buildFailureMessage());
```
rgabbard: Throw an exception on failure.  

(Crashes are much better than failures the user doesn't notice)

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/IndexGigawordWithElasticSearch.java
shinyichen: This is to check `filePath` is defined in the parameter. According to the documentation: Throws a ParameterException unless exactly one parameter is defined.

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/IndexGigawordWithElasticSearch.java
shinyichen: I totally missed it.

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/IndexGigawordWithElasticSearch.java
rgabbard: The use case for `assertExactlyOneDefined` is when you have, e.g. three parameters `a`, `b`, and `c` and you want to enforce that *exactly* one of them is defined.  A normal parameter access will fail if the parameter isn't defined, so there is no need to do this for a single parameter.

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/GigawordParserUtils.java
```diff
@@ -0,0 +1,126 @@
...
+   */
+  public static HashMap<String, String> getArticlesFromGZipFile(String filePath)
+      throws FileNotFoundException, IOException {
+    String fullText = gZipFileToString(filePath);
+    HashMap<String, String> articles = splitTextIntoArticles(fullText);
+    return articles;
+  }
+
+  public static void main(String[] args) {
+    String filePath = "/Users/jenniferchen/Downloads/gigaword_eng_5/data/afp_eng/afp_eng_199405.gz";
```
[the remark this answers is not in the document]
shinyichen: Fixed, using a input file.

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/GigawordParserUtils.java
```diff
@@ -0,0 +1,126 @@
...
+    try {
+      String fullText = gZipFileToString(filePath);
+      HashMap<String, String> articles = splitTextIntoArticles(fullText);
+      Iterator<String> itr = articles.keySet().iterator();
+      String docId1 = itr.next();
+      System.out.println(articles.get(docId1));
+    } catch (IOException e) {
+
+      e.printStackTrace();
```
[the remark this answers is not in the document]
shinyichen: Fixed.

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/IndexGigawordWithElasticSearch.java
shinyichen: Fixed by removing assert statements and catch MissingParameterException.

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/IndexGigawordWithElasticSearch.java
shinyichen: Fixed, please check.

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/IndexGigawordWithElasticSearch.java
shinyichen: Fixed.

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/IndexGigawordWithElasticSearch.java
```diff
@@ -1,52 +1,190 @@
...
+        for (File concatenatedFileToIndex : gigawordDir.listFiles()) {
+          if (concatenatedFileToIndex.isFile() && isValidGzipFile(concatenatedFileToIndex)) {
+            index(client, concatenatedFileToIndex, indexName);
+          }
         }
+      } else {
+        File concatenatedFileToIndex = parameters.getExistingFile(PARAM_GIGAWORD_FILEPATH);
+        if (!isValidGzipFile(concatenatedFileToIndex)) {
+        	client.close();
+        	throw new Exception(concatenatedFileToIndex.getAbsolutePath() + " is not a valid GZip file");
```
rgabbard: You want an `IOException` here - even if you want to throw an exception of unspecified type, prefer `RuntimeException` (because it is unchecked.  It doesn't matter much here in `main`, but it's good to be consistent)

rgabbard: Rather than sprinkling `client.close()` in various branches, put it in a `finally` block

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/IndexGigawordWithElasticSearch.java
```diff
@@ -1,52 +1,190 @@
...
+      for (Article article : articles) {
+        XContentBuilder sourceBuilder =
+            buildSourceObject(article, "en", "", new Date().toString(), "");
+        bulkRequest.add(
+            new IndexRequest(indexName, "texts", article.getId()).source(sourceBuilder));
+      }
+
+      BulkResponse bulkResponse = client.bulk(bulkRequest, RequestOptions.DEFAULT);
+      if (bulkResponse.hasFailures()) {
+        throw new Exception(bulkResponse.buildFailureMessage());
```
rgabbard: Switch to `RuntimeException`

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/IndexGigawordWithElasticSearch.java
```diff
@@ -1,52 +1,190 @@
...
+      }
+    }
+  }
+
+  private static XContentBuilder buildSourceObject(
```
rgabbard: Add a comment noting that this is the format expected for Inception's ElasticSearch-based external search, and point to the relevant part of the Inception code

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/IndexGigawordWithElasticSearch.java
shinyichen: Fixed. Please check.

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/IndexGigawordWithElasticSearch.java
shinyichen: Added.

PR #9, opened 2018-12-14
Index all gigaword

rgabbard:
Gigaword indexer can index all of Gigaword in its original directory layout


gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/IndexGigawordWithElasticSearch.java
```diff
@@ -62,84 +70,65 @@ public static void main(String[] argv) throws IOException {
...
-        index(client, concatenatedFileToIndex, indexName);
+        index(client, concatenatedFileToIndex.toPath(), indexName);
       }
 
     } catch (Exception e) {
       log.error("Caught exception: {}", e);
       System.exit(1);
-    } finally {
-      if (client != null) {
-        client.close();
```
shinyichen: Was it intentional to leave `client.close()` out?

rgabbard: It gets called automatically by the [try-with-resources ](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html) statement.

PR #14, opened 2019-02-15
Documentation and warnings for Gigaword version.

shinyichen:
* Added documentation for the accepted Gigaword version ID.
* Added warning when possible wrong Gigaword version detected.

### DIFF
--- a/gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/IndexGigawordWithElasticSearch.java
+++ b/gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/IndexGigawordWithElasticSearch.java
@@ -28,6 +28,7 @@ import static com.google.common.collect.Iterables.partition;
 
 /**
  * Indexes Gigaword with Elastic Search in a way usable by the external search feature of the Inception annotator.
+ * The expected Gigaword version is LDC2011T07 (plain version)
  *
  * See usage message for details.
  */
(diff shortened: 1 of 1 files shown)


gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/IndexGigawordWithElasticSearch.java
```diff
@@ -125,6 +127,12 @@ private static RestHighLevelClient buildElasticSearchClient(Parameters parameter
   private static void index(RestHighLevelClient client, Path file, String indexName) throws IOException {
     log.info("Indexing {}", file.toAbsolutePath());
 
+    // If file ends with .xml.gz, this may be the wrong version of Gigaword.
+    if (file.toString().toLowerCase().endsWith(".xml.gz")) {
```
rgabbard: This works, but `file.getName().toLowerCase()....` would be cleaner.

shinyichen: Both `getName(index)` and `getFileName()` return a `Path`.

rgabbard: Oh, it's a `Path` object, not a `File`.  `toString` is fine then

PR #19, opened 2019-03-12
Support for annotated Gigaword versions

az79nefy:
This PR contains a modification of gigaword-indexer to support annotated Gigaword versions.

Content of the PR:
- Choice of document extraction method after determination of Gigaword version
- Extraction of <HEADLINES> and <TEXT> passages from all documents of annotated Gigaword version 
- Parsing of extracted passages by DKPro Penntree and converting to plain text
- Creation of article objects
- Iteration over article objects by custom iterator


gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/ConcatenatedGigawordDocuments.java
```diff
@@ -18,18 +26,118 @@
...
+  }
+  
   public static ConcatenatedGigawordDocuments fromGigwordGZippedFile(Path p) throws IOException {
     return new ConcatenatedGigawordDocuments(GZIPByteSource.fromCompressed(p.toFile())
             .asCharSource(StandardCharsets.UTF_8).read());
   }
 
+  public static ConcatenatedGigawordDocuments fromAnnotatedGigwordGZippedFile(Path p) throws IOException {
+    // Initialize Streams
+    InputStream fileInputStream = new FileInputStream(p.toString());
```
reckart: Shouldn't this be done with a try-with-resources block to ensure that the streams are properly closed?

reckart:
@rgabbard We thought you might be interested in that :)

rgabbard:
@reckart : Thanks!  We can do the reverse of what we do in the Inception repository: when you are happy with your review pass, poke me and I'll review and merge :-)

az79nefy:
@reckart I added the try-with-resources block. Let me know if this is what you meant.

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/ConcatenatedGigawordDocuments.java
```diff
@@ -18,18 +26,112 @@
...
+                }
+                else {
+                    throw new RuntimeException("Missing document ID on article");
+                }
+            }
+    
+            // create article when end of document (</TEXT>) is reached
+            if (sCurrentLine.equals("</TEXT>")) {
+                articleList.add(new Article(currentDocId, docText));
+                docText = "";
```
reckart: @az79nefy You should use a StringBuilder instead of adding to the docText string using `+=`.

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/ConcatenatedGigawordDocuments.java
```diff
@@ -18,18 +26,112 @@
...
+                if (sCurrentLine.equals("")) {
+                    docText += "\n";
+                }
+                else if (sCurrentLine.startsWith("(")) {
+                    String parsedText = PennTreeUtils.toText(PennTreeUtils.parsePennTree(sCurrentLine));
+                    if (!parsedText.equals("null")) {
+                        docText += parsedText + " ";
+                    }
+                }
+                // sometimes there is one unannotated line in the document which can be used as plain text
```
reckart: When does this happen?

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/ConcatenatedGigawordDocuments.java
```diff
@@ -18,18 +26,112 @@
...
+                else {
+                    docText += sCurrentLine + " ";
```
reckart: Why add a space at the end of the line?

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/ConcatenatedGigawordDocuments.java
```diff
@@ -18,18 +26,112 @@
...
+                    currentDocId = m.group(1);
+                }
+                else {
+                    throw new RuntimeException("Missing document ID on article");
+                }
+            }
+    
+            // create article when end of document (</TEXT>) is reached
+            if (sCurrentLine.equals("</TEXT>")) {
+                articleList.add(new Article(currentDocId, docText));
```
reckart: Should probably trim the docText here to remove any trailing spaces?

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/ConcatenatedGigawordDocuments.java
```diff
@@ -18,18 +26,112 @@
...
+            if (sCurrentLine.equals("<TEXT>")) {
+                textStarted = true;
+            }
+    
+            if (sCurrentLine.equals("<HEADLINE>")) {
+                headlineStarted = true;
+            }
+        }
+    
+        return new ConcatenatedGigawordDocuments("", articleList);
```
reckart: Why is the empty string passed here as the first argument?

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/ConcatenatedGigawordDocuments.java
az79nefy: Because every line contains a sentence and there should be a space between the full stop of one sentence and the next sentence.

reckart: If every line has a sentence, then we already have a line-break between the full stop and the next sentence, right? I don't think it is necessary to add an additional space character before/after the line break.

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/ConcatenatedGigawordDocuments.java
```diff
@@ -18,18 +26,112 @@
...
+    this.articleList = aArticleList;
+    this.annotated = true;
+  }
+  
   public static ConcatenatedGigawordDocuments fromGigwordGZippedFile(Path p) throws IOException {
     return new ConcatenatedGigawordDocuments(GZIPByteSource.fromCompressed(p.toFile())
             .asCharSource(StandardCharsets.UTF_8).read());
   }
 
+  public static ConcatenatedGigawordDocuments fromAnnotatedGigwordGZippedFile(Path p) throws IOException {
```
rgabbard: I haven't yet done a full review pass, but wouldn't it be cleaner to do this as a separate class (e.g. `ConcatenatedAn notatedGigawordDocuments`) rather than within `ConcatenatedGigawordDocuments`?  The two code paths don't seem to share anything significant.

rgabbard: Also, is the code below trying to maintain a correspondence between the offsets of the returned document text and the original source Gigaword document?

reckart: @rgabbard As far as I could tell there is no way to do that. The Annotated Gigaword does not seem to include a "plain text" version of the data. The text comes in bracketed PTB format. We have no idea where whitespace was in the original Gigaword.

reckart: @az79nefy can you split it out into another class?

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/ConcatenatedGigawordDocuments.java
az79nefy: There is no line-break between the full stop and the next sentence currently. I can do that, but wouldn't I just replace the `+ " "` with `+ "\n"` then?

az79nefy: Currently line-breaks are only inserted if there is an empty line between to sentence-annotations. Which is only the case between paragraphs and not always.

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/ConcatenatedGigawordDocuments.java
rgabbard: @reckart : I don't have Annotated Gigaword downloaded, but from the sample of the LDC catalog it looks like offset information is stored on tokens using `CharacterOffsetBegin` and `CharacterOffsetEnd`:
LDC sample

Using those you could construct a text to index which corresponds to the original text offsets perfectly for all tokens, which is all you care about.  

The reason to do this is that otherwise annotations become difficult to distribute (since the original text cannot even be transferred from one institution to another, much less posted publicly). But if everything is in terms of the original offsets, everyone with an LDC membership can just download Gigaword and apply the annotations.

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/ConcatenatedGigawordDocuments.java
az79nefy: I really don't see a pattern. It almost seems as if some sentences just were forgotten.
Annotated sentences are usually between <P> and </P> tags, and these unannotated sentences are as well.

az79nefy: Here's an example:

```
<P>
( (S (NP (NNP Zito)) (ADVP (RB also)) (VP (VP (MD will) (VP (VB send) (PRT (RP out)) (NP (PRP It) (VBZ 's)) (PP (IN So) (NP (NP (NN Simple)) (PP (IN in) (NP (DT the) (NNP Wood) (NNP Memorial))) (PP (IN on) (NP (NNP Saturday))) (PP (IN at) (NP (NNP Aqueduct))))))) (, ,) (CC but) (VP (VBD said) (NP (NNP Wednesday)) (SBAR (S (NP (PRP he)) (VP (VBZ considers) (S (NP (DT A) (NNP P) (NNP Valentine)) (VP (TO to) (VP (VB be) (NP (PRP$ his) (JJ top) (NN 3-year-old)))))))))) (. .)))
</P>
<P>
A P Valentine 's rivals in the $ 750,000 Blue Grass are expected to be Hero 's Tribute , this year an allowance winner and a troubled third in the Louisiana Derby ; Dollar Bill , winner of the Risen Star and fourth in the Louisiana Derby after almost falling ; Invisible Ink , third in the Florida Derby ; Millennium Wind , the Santa Catalina winner and second in the Louisiana Derby ; Bonnie Scott , winner of four of six starts at Turfway Park ; and Songandaprayer , winner of the Fountain of Youth , then fifth in the Florida Derby .
</P>
<P>
( (S (NP (NNP Songandaprayer)) (VP (AUXZ is) (VP (VBN co-owned) (PP (IN by) (NP (NP (NNP Bobby) (NNP Hurley)) (, ,) (NP (NP (NP (DT an) (JJ All-American) (NN guard)) (PP (IN at) (NP (NNP Duke)))) (CC and) (NP (DT a) (JJ former) (NNP NBA) (NN player))))))) (. .)))
</P>
```

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/ConcatenatedGigawordDocuments.java
az79nefy: Fixed in newest commit.

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/ConcatenatedGigawordDocuments.java
az79nefy: Fixed in newest commit.

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/ConcatenatedGigawordDocuments.java
az79nefy: Fixed in newest commit.

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/ConcatenatedGigawordDocuments.java
az79nefy: Split `ConcatenatedAnnotatedGigawordDocument` into new class in new commit.

reckart:
@rgabbard I'd say for what it is meant to do and for our needs at least (namely being able to index AGW to query it from INCEpTION) the PR is ok. Thanks for the pointer regarding the offset information - I have no idea why we missed that. I agree that reconstructing the document text from this information would be much better. But I don't know if/when we'll be able to do that.

rgabbard:
@reckart : Got it.  I'm going to do a review with a few requests for things to make clear to users of this repo that the Annotated Gigaword support is for the convenience of another project and won't work with the rest of the "curated training" support pipeline we will be building here.  And of course, if it is less work for you all, feel free to fork the repo or copy the code useful to you elsewhere.

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/ConcatenatedAnnotatedGigawordDocuments.java
```diff
@@ -0,0 +1,133 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.zip.GZIPInputStream;
+
+/**
+ * The LDC distributes Gigaword as a moderate number of gzipped files, each of which has many documents
+ * concatenated together.  This class lets you iterate over the documents stored in such a file.
+ */
+
+public class ConcatenatedAnnotatedGigawordDocuments implements Iterable<ConcatenatedGigawordDocuments.Article> {
```
rgabbard: In the Javadoc, can you mention that:

* this class was authored by you all ("Technische Universität Darmstadt" or whatever the preferred nomenclature is)
* it does not preserve the offsets from the original text, so it shouldn't be used for the curated training process.
* it is included here for Darmstadt's convenience.

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/ConcatenatedGigawordDocuments.java
```diff
@@ -36,7 +36,7 @@ public static ConcatenatedGigawordDocuments fromGigwordGZippedFile(Path p) throw
     private String id;
     private String text;
 
-    private Article(String id, String text) {
```
rgabbard: Now that it is being used in two places (and soon we will be adding a third), can you promote this to a top-level class?

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/IndexGigawordWithElasticSearch.java
```diff
@@ -127,17 +127,24 @@ private static RestHighLevelClient buildElasticSearchClient(Parameters parameter
...
+      concatenatedGigawordDocuments = ConcatenatedGigawordDocuments.fromGigwordGZippedFile(file);
     }
 
     // we batch the documents in groups of 100 so we can get the efficiency gains from batching without
     // making huge requests of unbounded size
-    for (List<Article> articles : partition(ConcatenatedGigawordDocuments.fromGigwordGZippedFile(file), 100)) {
+    for (List<Article> articles : partition(concatenatedGigawordDocuments, 100)) {
       final BulkRequest bulkRequest = new BulkRequest();
       for (Article article : articles) {
+        System.out.println(article.getId());
```
rgabbard: This is going to lead to a lot of not very useful terminal output - we could either (a) remove it or (b) do something like print a "." every 10k document?

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/IndexGigawordWithElasticSearch.java
```diff
@@ -127,17 +127,24 @@ private static RestHighLevelClient buildElasticSearchClient(Parameters parameter
   private static void index(RestHighLevelClient client, Path file, String indexName) throws IOException {
     log.info("Indexing {}", file.toAbsolutePath());
 
+    Iterable<ConcatenatedGigawordDocuments.Article> concatenatedGigawordDocuments;
```
rgabbard: Can you add a string parameter `corpus_format` to the parameter file input to the program and have the decision about how to ingest the data come from that rather than the filename? That will be more robust, especially as we add more options. I'd suggest `gigaword` and`annotated_gigaword` as the currently-supported values.   If `annotated_gigaword` is chosen, please log a warning that offsets are not preserved and such a corpus cannot be used for the curated training process.

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/ConcatenatedAnnotatedGigawordDocuments.java
az79nefy: Fixed in newest commit.

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/ConcatenatedGigawordDocuments.java
az79nefy: Fixed in newest commit.

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/IndexGigawordWithElasticSearch.java
az79nefy: Removed terminal output in newest commit.

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/IndexGigawordWithElasticSearch.java
```diff
@@ -124,18 +126,27 @@ private static RestHighLevelClient buildElasticSearchClient(Parameters parameter
...
+      log.warn("Indexing an annotated version of Gigaword. Please note that this procedure does not preserve " +
+              "the offsets from the original document texts and therefore cannot be used for " +
+              "the curated training process.");
+      concatenatedGigawordDocuments = ConcatenatedAnnotatedGigawordDocuments.fromAnnotatedGigwordGZippedFile(file);
+    }
+    else if (corpusFormat.equals("gigaword")) {
+      concatenatedGigawordDocuments = ConcatenatedGigawordDocuments.fromGigwordGZippedFile(file);
+    }
+    else {
+      System.err.println(USAGE);
```
rgabbard: Can you put an explicit error here that the `corpusFormat` is unknown? Otherwise the user will not be sure what's wrong.

az79nefy:
@rgabbard I have followed your suggestion of using the offset information for the reconstruction of the original document texts.

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/IndexGigawordWithElasticSearch.java
az79nefy: Changed in latest commit.

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/AnnotatedGigawordParser.java
```diff
@@ -0,0 +1,92 @@
...
+    
+    @Override
+    public void endElement(String uri, String localName, String qName)
+    {
+        if (qName.equals("DOC")) {
+            inDocument = false;
+        }
+        else if (inDocument && qName.equals("sentences")) {
+            inSentences = false;
+            articleList.add(new Article(currentDocId, docText.toString().trim()));
```
rgabbard: You don't want to trim anything from the left or you will throw off your offsets if you happen to have leading whitespace (very unlikely to be in the first token, but not impossible).

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/AnnotatedGigawordParser.java
```diff
@@ -0,0 +1,92 @@
...
+            docText = new StringBuilder();
+        }
+        else if (inSentences && qName.equals("token")) {
+            inToken = false;
+            while (docText.toString().length() < currentOffsetBegin) {
```
rgabbard: Prefer simply `doctText.length()` - otherwise you need to copy the full document on every run through the while loop (unless their is some optimization which saves you).

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/AnnotatedGigawordParser.java
az79nefy: Changed in latest commit.

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/AnnotatedGigawordParser.java
az79nefy: Changed in latest commit.

rgabbard:
@az79nefy : this looks good now, but it has merge conflicts with `master` due to changes we've made in the meantime to support another format.  Can you fix the conflicts (via rebase is best, but if that is too messy you can take another approach) and then I'll merge?

az79nefy:
@rgabbard I have merged this branch with master.

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/IndexGigawordWithElasticSearch.java
```diff
@@ -38,8 +38,9 @@
           "\tindexName: the name of the index in a running Elastic Search server to add the documents to\n" +
           "\tcorpusDirectoryPath: the path to a directory where corpus (i.e. LDC2011T07 English Gigaword 5th edition)\n" +
           "\t\thas been extracted. \n" +
-          "\t\tIf indexing Gigaword, make sure the version is the plain Gigaword (LDC2011T07) file and not the other versions.\n" +
           "\tformat: LTF or gigaword" +
+          "\tcorpusFormat: flag whether the version if Gigaword is plain (\"gigaword\") or " +
```
rgabbard: I don't think we want two separate parameters - just make `annotated_gigaword` another option for `format`

az79nefy: Fixed in latest commit.

rgabbard: @az79nefy : The code is updated but the usage message still has two separate parameters.

rgabbard:
@az79nefy : Merged, thanks!

az79nefy:
:)

PR #33, opened 2019-06-12
JSON export and rehydration (issue 30)

qpwo:
Will add rehydration later

### DIFF
--- a/annotation-utils/pom.xml
+++ b/annotation-utils/pom.xml
@@ -94,6 +94,11 @@
 			<artifactId>common-core-open</artifactId>
 		</dependency>
 		<dependency>
+			<groupId>edu.isi.vista</groupId>
+			<artifactId>nlp-core-open</artifactId>
+            <version>8.2.2-SNAPSHOT</version>
+		</dependency>
+		<dependency>
 			<groupId>org.slf4j</groupId>
 			<artifactId>slf4j-api</artifactId>
 		</dependency>
@@ -143,7 +148,6 @@
 			<artifactId>jackson-core</artifactId>
 			<version>2.9.8</version>
 		</dependency>
-
 
 
 		<dependency>
--- a/annotation-utils/pom.xml
+++ b/annotation-utils/pom.xml
@@ -96,7 +96,7 @@
 		<dependency>
 			<groupId>edu.isi.vista</groupId>
 			<artifactId>nlp-core-open</artifactId>
-            <version>8.2.2-SNAPSHOT</version>
+            <version>8.2.2</version>
 		</dependency>
 		<dependency>
 			<groupId>org.slf4j</groupId>
--- a/annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExportAnnotations.kt
+++ b/annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExportAnnotations.kt
@@ -106,27 +106,32 @@ fun main(argv: Array<String>) {
 
             // an annotation record records user's annotate state for the document
             for (annotationRecord in annotationRecords) {
+                if (annotationRecord.state == "NEW") {
+                    // no actual annotation
+                    logger.warn { "Skipping $annotationRecord in $project because it is NEW" }
+                    continue
+                }
                 val getAnnotationsUrl = "$inceptionUrl/api/aero/v1/projects/${project.id}" +
                         "/documents/${document.id}/annotations/${annotationRecord.user}"
 
                 // the return from this will be the bytes of a a zip file which contains the
                 // JSON representation of the annotation
                 val annotationFileBytes =
-                    getAnnotationsUrl
-                            .httpGet(
-                                    parameters = listOf(
-                                            "projectId" to project.id,
-                                            "documentId" to document.id,
-                                            "userId" to annotationRecord.user,
-                                            // without this, it just downloads the original source
-                                            // text
-                                            "format" to "json"
-                                    )
-                            )
-                            .authenticateToInception()
-                            .retryOnFailure()
+                        getAnnotationsUrl
+                                .httpGet(
+                                        parameters = listOf(
+                                                "projectId" to project.id,
+                                                "documentId" to document.id,
+                                                "userId" to annotationRecord.user,
+                                                // without this, it just downloads the original source
+                                                // text
+                                                "format" to "json"
+                                        )
+                                )
+                                .authenticateToInception()
+                                .retryOnFailure()
                 if (annotationFileBytes == null) {
-                    logger.warn { "Skipping $annotationRecord in $project" }
+                    logger.warn { "Skipping $annotationRecord in $project due to timeout" }
                     continue
                 }
 
@@ -173,12 +178,13 @@ private data class Document(val id: Long, val name: String, val state: String) {
         }
     }
 }
+
 private data class DocumentResult(val messages: List<String>, val body: List<Document>)
 private data class AnnotatorRecord(val user: String, val state: String, val timestamp: String?)
 private data class AeroResult<T>(val messages: List<String>, val body: List<T>)
 
 private inline fun <reified T : Any> Request.resultObjectThrowingExceptionOnFailure(
-    mapper: ObjectMapper
+        mapper: ObjectMapper
 ): T {
     val (_, _, result) = this.responseObject<T>(mapper)
 
--- a/annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExportAnnotations.kt
+++ b/annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExportAnnotations.kt
@@ -3,6 +3,7 @@ package edu.isi.vista.annotationutils
 import com.fasterxml.jackson.databind.ObjectMapper
 import com.fasterxml.jackson.databind.node.ObjectNode
 import com.fasterxml.jackson.module.kotlin.registerKotlinModule
+import com.github.kittinunf.fuel.core.FuelError
 import com.github.kittinunf.fuel.core.Request
 import com.github.kittinunf.fuel.core.extensions.authentication
 import com.github.kittinunf.fuel.httpGet
@@ -41,6 +42,9 @@ val logger = KLogging().logger
  *     hierarchy beneath it will look like "projectName/documentName/documentName-userName.json"</li>
  *  </ul>
  */
+
+private data class Project(val id: Long, val name: String)
+
 fun main(argv: Array<String>) {
     if (argv.size != 1) {
         throw RuntimeException("Expected a single argument, a parameter file")
@@ -67,20 +71,31 @@ fun main(argv: Array<String>) {
     fun Request.authenticateToInception() =
             this.authentication().basic(inceptionUserName, inceptionPassword)
 
-    val projects = "$inceptionUrl/api/aero/v1/projects".httpGet()
-            .authenticateToInception()
-            .resultObjectThrowingExceptionOnFailure<AeroResult<Project>>(mapper)
-            .body
+    val projects = retryOnFuelError {
+        "$inceptionUrl/api/aero/v1/projects".httpGet()
+                .authenticateToInception()
+                .resultObjectThrowingExceptionOnFailure<AeroResult<Project>>(mapper)
+                .body
+    }
+    if (projects == null) {
+        throw java.lang.RuntimeException("Could not fetch projects from $inceptionUrl. Aborting")
+    }
     logger.info { "Projects on server ${projects.map { it.name }}" }
 
     for (project in projects) {
         logger.info { "Processing project $project" }
         val getDocsUrl = "$inceptionUrl/api/aero/v1/projects/${project.id}/documents"
-        val documents = getDocsUrl.httpGet(
-                parameters = listOf("projectId" to project.id))
-                .authenticateToInception()
-                .resultObjectThrowingExceptionOnFailure<DocumentResult>(mapper)
-                .body
+        val documents = retryOnFuelError {
+            getDocsUrl.httpGet(
+                    parameters = listOf("projectId" to project.id))
+                    .authenticateToInception()
+                    .resultObjectThrowingExceptionOnFailure<DocumentResult>(mapper)
+                    .body
+        }
+        if (documents == null) {
+            logger.warn { "Skipping $project due to web errors" }
+            continue
+        }
 
         // to reduce clutter, we only make a directory for a project if it in fact has any
         // annotation
@@ -93,22 +108,28 @@ fun main(argv: Array<String>) {
             // each annotator's annotation for a document are stored separately
             val getAnnotatingUsersUrl = "$inceptionUrl/api/aero/v1/projects/" +
                     "${project.id}/documents/${document.id}/annotations"
-            val annotationRecords = getAnnotatingUsersUrl
-                    .httpGet(
-                            parameters = listOf(
-                                    "projectId" to project.id,
-                                    "documentId" to document.id
-                            )
-                    )
-                    .authenticateToInception()
-                    .resultObjectThrowingExceptionOnFailure<AeroResult<AnnotatorRecord>>(mapper)
-                    .body
+            val annotationRecords = retryOnFuelError {
+                getAnnotatingUsersUrl
+                        .httpGet(
+                                parameters = listOf(
+                                        "projectId" to project.id,
+                                        "documentId" to document.id
+                                )
+                        )
+                        .authenticateToInception()
+                        .resultObjectThrowingExceptionOnFailure<AeroResult<AnnotatorRecord>>(mapper)
+                        .body
+            }
+            if (annotationRecords == null) {
+                logger.warn { "Skipping $document due to network errors" }
+                continue
+            }
 
             // an annotation record records user's annotate state for the document
             for (annotationRecord in annotationRecords) {
                 if (annotationRecord.state == "NEW") {
                     // no actual annotation
-                    logger.warn { "Skipping $annotationRecord in $project because it is NEW" }
+                    logger.warn { "Skipping $annotationRecord because it is NEW" }
                     continue
                 }
                 val getAnnotationsUrl = "$inceptionUrl/api/aero/v1/projects/${project.id}" +
@@ -129,9 +150,9 @@ fun main(argv: Array<String>) {
                                         )
                                 )
                                 .authenticateToInception()
-                                .retryOnFailure()
+                                .retryOnResponseFailure()
                 if (annotationFileBytes == null) {
-                    logger.warn { "Skipping $annotationRecord in $project due to network error" }
+                    logger.warn { "Skipping $annotationRecord due to network errors" }
                     continue
                 }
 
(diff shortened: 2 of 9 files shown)


annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExportAnnotations.kt
```diff
@@ -150,11 +151,12 @@ fun main(argv: Array<String>) {
...
-                    val jsonBytes = it.getInputStream("$zipEntryName.json")?.readAllBytes()
-
+                    val jsonBytes = it.getInputStream("$zipEntryName.json")?.readBytes()
                     if (jsonBytes != null) {
-                        val outFileName = documentOutputDir.resolve("${document.name}-${annotationRecord.user}.json");
-                        Files.write(outFileName, jsonBytes)
+                        var jsonTree = ObjectMapper().readTree(jsonBytes) as ObjectNode
+                        jsonTree.replaceFieldEverywhere("sofaString", "__DOCUMENT_TEXT_REDACTED_FOR_IP_REASONS__")
+                        val outFileName = documentOutputDir.resolve("${document.name}-${annotationRecord.user}.json")
+                        Files.write(outFileName, jsonTree.toString().toByteArray())
```
rgabbard: In general, `toString` is to be avoided except for diagnostic purposes. Prefer to generate the JSON string using `ObjectMapper`.

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExportAnnotations.kt
```diff
@@ -150,11 +151,12 @@ fun main(argv: Array<String>) {
...
                     }
 
-                    val jsonBytes = it.getInputStream("$zipEntryName.json")?.readAllBytes()
-
+                    val jsonBytes = it.getInputStream("$zipEntryName.json")?.readBytes()
                     if (jsonBytes != null) {
-                        val outFileName = documentOutputDir.resolve("${document.name}-${annotationRecord.user}.json");
-                        Files.write(outFileName, jsonBytes)
+                        var jsonTree = ObjectMapper().readTree(jsonBytes) as ObjectNode
+                        jsonTree.replaceFieldEverywhere("sofaString", "__DOCUMENT_TEXT_REDACTED_FOR_IP_REASONS__")
```
rgabbard: Add a comment about why we do this (that we cannot redistribute LDC documents and the user can always restore from those LDC documents)

pom.xml
```diff
@@ -63,7 +63,7 @@
 			<dependency>
 				<groupId>edu.isi.vista</groupId>
 				<artifactId>common-core-open</artifactId>
-				<version>8.2.1-SNAPSHOT</version>
+				<version>8.2.2-SNAPSHOT</version>
```
rgabbard: What are we depending on which is not present in the release version?

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExportAnnotations.kt
```diff
@@ -100,38 +125,27 @@ fun main(argv: Array<String>) {
                     .resultObjectThrowingExceptionOnFailure<AeroResult<AnnotatorRecord>>(mapper)
                     .body
 
-            val documentOutputDir = projectOutputDir.resolve(document.name)
```
rgabbard: Why was this removed?

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExportAnnotations.kt
```diff
@@ -172,21 +190,37 @@ private data class Document(val id: Long, val name: String, val state: String) {
...
+    if (maxTries == 0)
+        throw RuntimeException("HTTP request has failed. Aborting.")
+    return when (result) {
+        is Result.Success<*> -> result.get()
+        is Result.Failure<*> -> {
+            logger.warn {
+                "Request timed out. Waiting $timeoutInSeconds seconds and trying again."
+            }
+            TimeUnit.SECONDS.sleep(timeoutInSeconds)
+            this.retryOnFailure(maxTries - 1, timeoutInSeconds)
```
rgabbard: In this case, I think the simple for loop is clearer than recursion

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExportAnnotations.kt
```diff
@@ -172,21 +190,37 @@ private data class Document(val id: Long, val name: String, val state: String) {
...
             result.get()
         }
     }
-}
\ No newline at end of file
+}
+
+private fun Request.retryOnFailure(maxTries: Int = 3, timeoutInSeconds: Long = 30): ByteArray {
+    val (_, _, result) = this.response()
+    if (maxTries == 0)
```
rgabbard: Always use braces

annotation-utils/src/main/java/edu/isi/vista/annotationutils/Rehydrate.kt
```diff
@@ -0,0 +1,75 @@
...
+import edu.isi.nlp.io.OffsetIndexedCorpus
+import edu.isi.nlp.symbols.Symbol
+import edu.isi.nlp.io.DocIDToFileMappings
+import com.google.common.base.Optional
+import com.fasterxml.jackson.databind.ObjectMapper
+import edu.isi.nlp.parameters.Parameters
+import java.io.File
+
+/**
+ * Puts the document text back into the 'sofaString' field inside json files created by `ExportAnnotations.kt`
```
rgabbard: Explain where and why they were removed

rgabbard: Also mention this program in the repo's `README.md`

rgabbard: Also, while we use the "dehydrate/rehydrate" language informally, it would be clearer to future readers if we called this something like `RestoreDocumentText`

annotation-utils/src/main/java/edu/isi/vista/annotationutils/Rehydrate.kt
```diff
@@ -0,0 +1,75 @@
...
+ */
+
+fun main(argv: Array<String>) {
+    // Load parameters:
+    val params = Parameters.loadSerifStyle(File(argv[0]))
+    val indexDirectory = params.getExistingDirectory("indexDirectory")
+    val gigawordDataDirectory = params.getExistingDirectory("gigawordDataDirectory")
+    val jsonInputDirectory = params.getCreatableDirectory("dehydratedJsonDirectory")
+    val outputDirectory = params.getCreatableDirectory("rehydratedJsonDirectory")
+    outputDirectory.toString()
```
rgabbard: ?

annotation-utils/src/main/java/edu/isi/vista/annotationutils/Rehydrate.kt
```diff
@@ -0,0 +1,75 @@
...
+ *     <li> {@code indexDirectory} is the location of the files produced by {@code IndexFlatGigaword.java}
+            (https://github.com/isi-vista/nlp-util/blob/master/nlp-core-open/src/main/java/edu/isi/nlp/corpora/gigaword/IndexFlatGigaword.java) </li>
+ *     <li> {@code gigawordDataDirectory} is the location of the gigaword text files </li>
+ *     <li> {@code jsonInputDirectory} is the location of the 'dehydrated' json files produced by {@code ExportAnnotations.kt} </li>
+ *     <li> {@code rehydratedJsonDirectory} is where the new json files will go </li>
+ *  </ul>
+ */
+
+fun main(argv: Array<String>) {
+    // Load parameters:
```
rgabbard: This comment doesn't add much :-)

annotation-utils/src/main/java/edu/isi/vista/annotationutils/Rehydrate.kt
```diff
@@ -0,0 +1,75 @@
...
+    val params = Parameters.loadSerifStyle(File(argv[0]))
+    val indexDirectory = params.getExistingDirectory("indexDirectory")
+    val gigawordDataDirectory = params.getExistingDirectory("gigawordDataDirectory")
+    val jsonInputDirectory = params.getCreatableDirectory("dehydratedJsonDirectory")
+    val outputDirectory = params.getCreatableDirectory("rehydratedJsonDirectory")
+    outputDirectory.toString()
+
+    // Create an OriginalTextSource for getting original document text to put in json:
+    val textMap = DocIDToFileMappings.forFunction { symbol ->
```
rgabbard: You could consider pulling everything from here through lines 44 into its own method (possibly just taking the `Parameters` as an argument)

annotation-utils/src/main/java/edu/isi/vista/annotationutils/Rehydrate.kt
```diff
@@ -0,0 +1,75 @@
...
+            continue
+        }
+        val projectOutDir = File(outputDirectory, projectDir.name)
+        projectOutDir.mkdir()
+        for (jsonFile in projectDir.listFiles()) {
+            val jsonTree = objectMapper.readTree(jsonFile)
+            // First 21 characters of the filename are the document id
+            // Example filename is AFP_ENG_19960918.0012-admin.json
+            val docID = Symbol.from(jsonFile.name.substring(0, 21))
+            val text = textSource.getOriginalText(docID).get()
```
rgabbard: Check and raise an exception when `getOriginalText` returns `Optional.absent`. A possible idiom for this is:

```
val text = textSource.getOriginalText(docId).orNull() ?: 
       throw RuntimeException("Could not get original text for $docId")
```

annotation-utils/src/main/java/edu/isi/vista/annotationutils/Rehydrate.kt
```diff
@@ -0,0 +1,75 @@
...
+            jsonTree.replaceFieldEverywhere("sofaString", text)
+            val outFile = File(projectOutDir, jsonFile.name)
+            outFile.writeBytes(prettyPrinter.writeValueAsBytes(jsonTree))
+        }
+    }
+}
+
+fun docIDToFilename(symbol: Symbol?): String? {
```
rgabbard: Why do we accept `Symbol?` here rather than requiring `Symbol`?

annotation-utils/pom.xml
```diff
@@ -93,6 +93,11 @@
 			<groupId>edu.isi.vista</groupId>
 			<artifactId>common-core-open</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>edu.isi.vista</groupId>
+			<artifactId>nlp-core-open</artifactId>
+            <version>8.2.2-SNAPSHOT</version>
```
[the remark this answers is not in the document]
qpwo: Deferring to @berquist here

annotation-utils/src/main/java/edu/isi/vista/annotationutils/Rehydrate.kt
qpwo: `DocIDToFileMappings.forFunction` sends a `Symbol?` but I made `docIDToFilename` throw an error

annotation-utils/pom.xml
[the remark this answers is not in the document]
berquist: I think this was done in the struggle to get things compiling, where the crucial step was adding the dependency to the inner pom, not specifically making it be `-SNAPSHOT`. @qpwo Can you try making all of the `-SNAPSHOT`s into the previous stable version and see if it still works?

berquist:
I think my commit https://github.com/isi-vista/curated-training-annotator/pull/33/commits/16bf9e11325879e5402f0552eaa31875cdca66b3 can be reverted (restore the deleted retry function and delete the recursive extension method) since we discovered the problem is with a particular annotation account on the server.

pom.xml
```diff
@@ -63,12 +63,12 @@
...
 			</dependency>
 			<dependency>
 				<groupId>edu.isi.vista</groupId>
 				<artifactId>nlp-core-open</artifactId>
-				<version>8.1.0</version>
+				<version>8.2.2-SNAPSHOT</version>
```
berquist: Try depending on `8.2.1`.

pom.xml
```diff
@@ -63,12 +63,12 @@
 			<dependency>
 				<groupId>edu.isi.vista</groupId>
 				<artifactId>common-core-open</artifactId>
-				<version>8.2.1-SNAPSHOT</version>
+				<version>8.2.2-SNAPSHOT</version>
```
berquist: Try depending on `8.2.1`.

annotation-utils/src/main/java/edu/isi/vista/annotationutils/Rehydrate.kt
```diff
@@ -0,0 +1,81 @@
...
+            val outFile = File(projectOutDir, jsonFile.name)
+            outFile.writeBytes(prettyPrinter.writeValueAsBytes(jsonTree))
+        }
+    }
+}
+
+fun makeTextSource(params: Parameters): OriginalTextSource {
+    // Create an OriginalTextSource for getting original document text to put in json:
+    val textMap = DocIDToFileMappings.forFunction { symbol ->
+        Optional.of(File(params.getExistingDirectory("gigawordDataDirectory"), docIDToFilename(symbol)))
```
berquist: Why is this `Optional`?

qpwo: `DocIDToFileMappings.forFunction` wouldn't take it non-optionally and I couldn't get an `as`-cast to work either

rgabbard: It expects a function which returns `Optional` because in general "I can't map this" is a valid response from a mapping strategy.

pom.xml
qpwo: Couldn't get it to work with 8.2.2. IntelliJ couldn't find it

rgabbard: Does it work for 8.2.1?

rgabbard: It looks like changes since 8.2.1 pertain only to `IndexFlatGigaword`: https://github.com/isi-vista/nlp-util/commits/master

qpwo: Oops! I couldn't get it to work for 8.2.**1** but it works for 8.2.**2** so I think we should keep these changes.

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExportAnnotations.kt
qpwo: The output structure had one directory per file so we flattened it into just one directory per project. Should we revert that?

rgabbard: How do we handle multiple annotators annotating the same file on the same project?

qpwo: Oh is that possible? I thought each user has their own fresh copy of the project? (E.g. named `ArtifactExistence.DamageDestroy-bob_loblaw`)

qpwo:
Did some more testing and indeed `8.2.2-SNAPSHOT` is required

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExportAnnotations.kt
rgabbard: By policy, yes, but it is possible in the Inception data model for the someone else to come in an annotate on the project with someone else's name on it, so we should handle that case.

qpwo: The containing dir names the project owner, and the file itself names the annotator. Does that eliminate possibility of collision? For example there is `output/ArtifactExistence.DamageDestroy-gabbard/AFP_ENG_19960918.0012-admin.json`

rgabbard: Yep, that will do the trick. Maybe add a comment somewhere to clarify this?

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExportAnnotations.kt
```diff
@@ -58,7 +58,9 @@ fun main(argv: Array<String>) {
     }
     logger.info { "Connecting to Inception at $inceptionUrl" }
 
-    val mapper = ObjectMapper().registerKotlinModule()
+    val objectMapper = ObjectMapper()
```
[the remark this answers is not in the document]
rgabbard: Can you return this to its original form?  It is more idiomatic Jackson code; otherwise the reader is going to say "why did they do that?"

annotation-utils/src/main/java/edu/isi/vista/annotationutils/Rehydrate.kt
rgabbard: Still need to rename away from re-hydrate

annotation-utils/pom.xml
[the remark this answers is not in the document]
rgabbard: 8.2.2 has now been released, so you should be able to use 8.2.2 (without `-SNAPSHOT`) as the value now

pom.xml
rgabbard: 8.2.2 here as well, and below

qpwo:
I get this error when I use 8.2.2 without the SNAPSHOT and run `ExportAnnotations.kt`:
```
10:32:32.808 [main] WARN  mu.KLogging - com.github.kittinunf.fuel.core.FuelError: Cannot construct instance of `edu.isi.vista.annotationutils.Document` (no Creators, like default construct, exist): cannot deserialize from Object value (no delegate- or property-based Creator)
 at [Source: (BufferedInputStream); line: 1, column: 25] (through reference chain: edu.isi.vista.annotationutils.DocumentResult["body"]->java.util.ArrayList[0])
	com.github.kittinunf.fuel.core.FuelError$Companion.wrap(FuelError.kt:85)
	com.github.kittinunf.fuel.core.DeserializableKt.response(Deserializable.kt:176)
```

I think this statement is causing it:
```
1.        val documents = retryOnFuelError {
2.            getDocsUrl.httpGet(
3.                    parameters = listOf("projectId" to project.id))
4.                    .authenticateToInception()
5.                    .resultObjectThrowingExceptionOnFailure<DocumentResult>(mapper)
6.                    .body
7.        }
```

Where line 5 is the problem.

But why wouldn't this happen on the SNAPSHOT version? :thinking:

rgabbard:
@qpwo : Are you sure the version of the `nlp-utils` dependencies reliably causes this difference?  If you look at  https://github.com/isi-vista/nlp-util/commits/master, you'll see that (unless you have local modifications to `nlp-utils`):

* `8.2.2` and `8.2.2-SNAPSHOT` are identical
* `8.2.1` and `8.2.2` differ only in code internal to `IndexFlatGigaword`, which should be irrelevant here
* `8.2.1` differs from `8.2.0` only in code internal to `Parameters`, which should be irrelevant here.\\

qpwo:
You're right; I changed it back to `8.2.2-SNAPSHOT` and I got the same error as before. I changed it back to `8.2.2` and then did a `mvn clean install` and that killed the error somehow. Some sort of problem with changing dependencies in IntelliJ perhaps?

rgabbard:
@qpwo : Had you edited the dependencies in IntelliJ's interface directly or only via the `pom.xml`?

qpwo:
`pom.xml`

rgabbard:
@qpwo : If the problem is not present on a clean build, let's not worry about it for now.

qpwo:
Yes I agree I committed the changes earlier

rgabbard:
@qpwo : Is this PR ready for re-review?

qpwo:
Yes!

berquist:
I'm now getting lots of `FuelError` messages, and things basically worked before except for the one problem with https://github.com/isi-vista/curated-training-annotator/issues/34. Something happened between https://github.com/isi-vista/curated-training-annotator/pull/33/commits/5ea83801d0345d3e173cba5a067370e550bedd3b and the PR head.

rgabbard:
@qpwo : Poke me for re-review once you and @berquist have sorted out the problem above

qpwo:
@berquist Are you still getting the error after a `mvn clean install` and then `invalidate caches and restart` in IntelliJ?

qpwo:
Might be ready for a merge I'm still not getting errors

PR #37, opened 2019-08-22
adds code for pushing new annotation data to curated-training-annotat…

elizlee:
…ion repo

### DIFF
--- a/annotation-utils/pom.xml
+++ b/annotation-utils/pom.xml
@@ -234,7 +234,6 @@
 		<dependency>
 			<groupId>org.eclipse.jgit</groupId>
 			<artifactId>org.eclipse.jgit</artifactId>
-			<version>5.4.0.201906121030-r</version>
 		</dependency>
 
 
--- a/annotation-utils/pom.xml
+++ b/annotation-utils/pom.xml
@@ -230,6 +230,13 @@
 			<version>0.12.0</version>
 		</dependency>
 
+		<!-- for managing git repositories -->
+		<dependency>
+			<groupId>org.eclipse.jgit</groupId>
+			<artifactId>org.eclipse.jgit</artifactId>
+			<version>5.4.0.201906121030-r</version>
+		</dependency>
+
 
 	</dependencies>
 
--- a/annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
+++ b/annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
@@ -1,0 +1,145 @@
+package edu.isi.vista.annotationutils
+
+import com.jcraft.jsch.JSch
+import com.jcraft.jsch.Session
+import edu.isi.nlp.parameters.serifstyle.SerifStyleParameterFileLoader
+import java.io.File
+import org.eclipse.jgit.api.Git
+import org.eclipse.jgit.api.ResetCommand
+import org.eclipse.jgit.api.TransportConfigCallback
+import org.eclipse.jgit.diff.DiffEntry
+import org.eclipse.jgit.lib.ObjectId
+import org.eclipse.jgit.revwalk.RevCommit
+import org.eclipse.jgit.revwalk.RevWalk
+import org.eclipse.jgit.transport.*
+import org.eclipse.jgit.treewalk.CanonicalTreeParser
+import org.eclipse.jgit.util.FS
+import java.text.SimpleDateFormat
+import java.util.*
+
+/**
+ * The goal of this program is to export the annotations, save the restored
+ * documents, compare with the previous commit to see if any changes have been
+ * made, and if so, push those changes to the repo.
+ *
+ * The program takes one argument: a file containing these parameter values:
+ *
+ * <ul>
+ *     <li> {@code exportAnnotationsParams}: the location of the parameter file meant for {@code ExportAnnotations.kt} </li>
+ *     <li> {@code restoreJsonParams}: the location of the parameter file meant for {@code RestoreJson.kt} </li>
+ *     <li> {@code repoPath}: the location of the local repository
+ *     (https://github.com/isi-vista/curated-training-annotation)</li>
+ * </ul>
+ *
+ * This program currently assumes these things:
+ * <ul>
+ *     <li>The annotations will be saved to .../curated-training-annotation/data/exported</li>
+ *     <li>Your ssh private key is saved in ~/.ssh/id_rsa.</li>
+ * </ul>
+ */
+
+
+fun pushNew(repoPath: String) {
+
+    // Source: https://github.com/kittinunf/Kotlin-Playground/blob/master/buildSrc/src/main/java/com/github/kittinunf/gradle/ReleaseToMasterMergeTask.kt
+    class SshConfig(val sshKeyPath: String): TransportConfigCallback {
+        override fun configure(transport: Transport?) {
+            (transport as SshTransport).sshSessionFactory = object: JschConfigSessionFactory() {
+                override fun configure(hc: OpenSshConfig.Host?, session: Session?) {}
+                override fun createDefaultJSch(fs: FS?): JSch = super.createDefaultJSch(fs).apply {
+                    addIdentity(sshKeyPath)
+
+                }
+            }
+        }
+    }
+
+    val git = Git.open(File(repoPath))
+    val config = git.getRepository().getConfig()
+    config.setString(
+            "remote",
+            "origin",
+            "url",
+            "git@github.com:isi-vista/curated-training-annotation.git")
+    config.save()
+
+    val currentDate = Calendar.getInstance(TimeZone.getTimeZone("UTC"))
+    TimeZone.setDefault(TimeZone.getTimeZone("UTC"))
+    val formatter = SimpleDateFormat("yyyy-MM-dd-HH-mm-ss")
+
+    // Currently working on test_branch
+    git.checkout().setName("test_branch").call()
+
+    git.add().addFilepattern("data/").call()
+    val newCommit = git.commit()
+            .setMessage("includes annotations from ${currentDate.getTime()}").call()
+    git.tag().setName(formatter.format(currentDate.getTime())).call()
+
+
+    if (hasDiff(git, newCommit)) {
+        val pushIter = git.push()
+                .setTransportConfigCallback(SshConfig("~/.ssh/id_rsa"))
+                .call()
+        val pushResult = pushIter.iterator().next()
+    } else {
+        println("No new annotations - nothing will be pushed.")
+        git.reset().setRef("HEAD~1").setMode(ResetCommand.ResetType.SOFT).call()
+    }
+}
+
+// Compare old data to newly exported data
+fun hasDiff(git: Git, newCommit: RevCommit): Boolean {
+    val oldCommit = getPrevCommit(git, newCommit)
+    if (oldCommit == null) {
+        println("This is the first commit.")
+        return true
+    }
+    val oldTree = getCanonicalTreeParser(git, oldCommit)
+    val newTree = getCanonicalTreeParser(git, newCommit)
+    val diffs: List<DiffEntry> =
+            git.diff().setNewTree(newTree).setOldTree(oldTree).call()
+    if (diffs.isEmpty()) {
+        return false
+    }
+    return true
+}
+
+fun getPrevCommit(git: Git, commit: RevCommit): RevCommit? {
+    val walk = RevWalk(git.getRepository())
+    walk.markStart(commit)
+    var count = 0
+    for (rev in walk) {
+        if (count ==1) {
+            return rev
+        }
+        count = count.inc()
+    }
+    return null
+}
+
+fun getCanonicalTreeParser(git: Git, commit: ObjectId): CanonicalTreeParser {
+    val walk = RevWalk(git.getRepository())
+    val revCommit = walk.parseCommit(commit)
+    val treeId = revCommit.getTree().getId()
+    val reader = git.getRepository().newObjectReader()
+    return CanonicalTreeParser(null, reader, treeId)
+}
+
+
+fun main(argv: Array<String>) {
+    val paramsLoader = SerifStyleParameterFileLoader.Builder()
+            .interpolateEnvironmentalVariables(true).build()
+    val params = paramsLoader.load(File(argv[0]))
+    val exportAnnotationsParams = params.getString("exportAnnotationsParams")
+    val restoreJsonParams = params.getString("restoreJsonParams")
+    val repoPath = params.getString("repoPath")
+
+    // Run ExportAnnotations.kt
+    ExportAnnotations.main(arrayOf(exportAnnotationsParams))
+
+    // Run RestoreJson.kt
+    RestoreJson.main(arrayOf(restoreJsonParams))
+
+    // Push new annotations
+    pushNew(repoPath)
+}
--- a/annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
+++ b/annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
@@ -4,7 +4,6 @@ import com.jcraft.jsch.JSch
 import com.jcraft.jsch.Session
 import edu.isi.nlp.parameters.Parameters
 import edu.isi.nlp.parameters.serifstyle.SerifStyleParameterFileLoader
-import mu.KLogging
 import java.io.File
 import org.eclipse.jgit.api.Git
 import org.eclipse.jgit.api.ResetCommand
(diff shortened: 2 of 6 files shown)


annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
```diff
@@ -0,0 +1,145 @@
...
+import org.eclipse.jgit.lib.ObjectId
+import org.eclipse.jgit.revwalk.RevCommit
+import org.eclipse.jgit.revwalk.RevWalk
+import org.eclipse.jgit.transport.*
+import org.eclipse.jgit.treewalk.CanonicalTreeParser
+import org.eclipse.jgit.util.FS
+import java.text.SimpleDateFormat
+import java.util.*
+
+/**
```
rgabbard: 💯 for a good "how to use this" comment.

I suggest shifting the `main` method from the bottom of the file to the top and making this the docstring for that method.  (Generally, it is good to order the classes, functions, etc. in a source file in the order a user trying to understand the program should ready them. You have more freedom to do this in Java/Kotlin/etc. than in Python, which has some ordering constraints)

annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
```diff
@@ -0,0 +1,145 @@
...
+ * The goal of this program is to export the annotations, save the restored
+ * documents, compare with the previous commit to see if any changes have been
+ * made, and if so, push those changes to the repo.
```
rgabbard: In documentation, you always want to be careful with your pragmatics. When you say "the repo", it is clear to you and me, because we currently have all the relevant context, but a new reader might be confused.  A better option might be "push those changes to the Git repository specified by the `repoPath` parameter".

annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
```diff
@@ -0,0 +1,145 @@
...
+ *
+ * The program takes one argument: a file containing these parameter values:
+ *
+ * <ul>
+ *     <li> {@code exportAnnotationsParams}: the location of the parameter file meant for {@code ExportAnnotations.kt} </li>
```
rgabbard: One advantage of Kotlin's doc tool over Javascript is that you can use regular Markdown, so *I think* you can just surround these with backticks rather than needing the `@code`s.

annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
```diff
@@ -0,0 +1,145 @@
...
+ *     <li> {@code restoreJsonParams}: the location of the parameter file meant for {@code RestoreJson.kt} </li>
+ *     <li> {@code repoPath}: the location of the local repository
+ *     (https://github.com/isi-vista/curated-training-annotation)</li>
```
rgabbard: Maybe say here  "For ISI's event curated training work, this is https://github.com/isi-vista/curated-training-annotation but it will differ for other projects"

annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
```diff
@@ -0,0 +1,145 @@
...
+fun getCanonicalTreeParser(git: Git, commit: ObjectId): CanonicalTreeParser {
+    val walk = RevWalk(git.getRepository())
+    val revCommit = walk.parseCommit(commit)
+    val treeId = revCommit.getTree().getId()
+    val reader = git.getRepository().newObjectReader()
+    return CanonicalTreeParser(null, reader, treeId)
+}
+
+
+fun main(argv: Array<String>) {
```
rgabbard: As noted above, put this function first so the reader can get a high-level overview of what the program is doing before diving into details

annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
```diff
@@ -0,0 +1,145 @@
...
+    val paramsLoader = SerifStyleParameterFileLoader.Builder()
+            .interpolateEnvironmentalVariables(true).build()
+    val params = paramsLoader.load(File(argv[0]))
+    val exportAnnotationsParams = params.getString("exportAnnotationsParams")
```
rgabbard: Prefer `getExistingFile`, which will throw an exception immediately if the file does not exist.  Ditto for `restoreJsonParams` below

annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
```diff
@@ -0,0 +1,145 @@
...
+    val restoreJsonParams = params.getString("restoreJsonParams")
+    val repoPath = params.getString("repoPath")
```
rgabbard: A more explicit name like `repositoryToPushTo` makes the function of the repository clearer to the user.

annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
```diff
@@ -0,0 +1,145 @@
...
+ *
+ * This program currently assumes these things:
+ * <ul>
+ *     <li>The annotations will be saved to .../curated-training-annotation/data/exported</li>
+ *     <li>Your ssh private key is saved in ~/.ssh/id_rsa.</li>
+ * </ul>
+ */
+
+
+fun pushNew(repoPath: String) {
```
rgabbard: The name `pushNew` raises the question - new whats?  A better name might be `pushUpdatedAnnotations`

annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
```diff
@@ -0,0 +1,145 @@
...
+
+    // Source: https://github.com/kittinunf/Kotlin-Playground/blob/master/buildSrc/src/main/java/com/github/kittinunf/gradle/ReleaseToMasterMergeTask.kt
+    class SshConfig(val sshKeyPath: String): TransportConfigCallback {
```
rgabbard: 💯 for using a local class to limit it to the scope where it is used

annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
```diff
@@ -0,0 +1,145 @@
...
+                override fun configure(hc: OpenSshConfig.Host?, session: Session?) {}
+                override fun createDefaultJSch(fs: FS?): JSch = super.createDefaultJSch(fs).apply {
+                    addIdentity(sshKeyPath)
+
+                }
+            }
+        }
+    }
+
+    val git = Git.open(File(repoPath))
```
rgabbard: So, up to this point in reading this, I had assumed that `repoPath` was the URL of the remote repository you want to push to, but now it appears to be the path to the local working copy of that repository.  We should clarify this in the documentation.

However, in the interest of being self-contained, I'd suggest making a slight change to the design:
* have two parameters `localWorkingCopyDirectory` and `repositoryToPushTo`.
* if `localWorkingCopyDirectory` does not exist, clone `repositoryToPushTo` to that location.
* if it does exist, make sure it is in the expected state
     * check that its remote repository URL matches `repositoryToPushTo`.
     * check if has no local modifications (`status.isClean())`
     * check it is on the master branch
     * sync it up to the remote master

annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
```diff
@@ -0,0 +1,145 @@
...
+            "url",
+            "git@github.com:isi-vista/curated-training-annotation.git")
+    config.save()
+
+    val currentDate = Calendar.getInstance(TimeZone.getTimeZone("UTC"))
+    TimeZone.setDefault(TimeZone.getTimeZone("UTC"))
+    val formatter = SimpleDateFormat("yyyy-MM-dd-HH-mm-ss")
+
+    // Currently working on test_branch
+    git.checkout().setName("test_branch").call()
```
rgabbard: For the real thing, we won't need `test_branch` - we can push directly to master.  For testing, rather than using a branch on the real repository, you can make another throwaway repo on GitHub.

annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
```diff
@@ -0,0 +1,145 @@
...
+    git.tag().setName(formatter.format(currentDate.getTime())).call()
+
+
+    if (hasDiff(git, newCommit)) {
+        val pushIter = git.push()
+                .setTransportConfigCallback(SshConfig("~/.ssh/id_rsa"))
+                .call()
+        val pushResult = pushIter.iterator().next()
+    } else {
+        println("No new annotations - nothing will be pushed.")
```
rgabbard: Prefer to use proper logging over `println` (at the `INFO` level in this case).  For Kotlin, I suggest https://github.com/MicroUtils/kotlin-logging .

annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
```diff
@@ -0,0 +1,145 @@
...
+    val newCommit = git.commit()
+            .setMessage("includes annotations from ${currentDate.getTime()}").call()
+    git.tag().setName(formatter.format(currentDate.getTime())).call()
+
+
+    if (hasDiff(git, newCommit)) {
+        val pushIter = git.push()
+                .setTransportConfigCallback(SshConfig("~/.ssh/id_rsa"))
+                .call()
+        val pushResult = pushIter.iterator().next()
```
rgabbard: What would it mean for there to be multiple push results? Is this something we need to be concerned with?

annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
```diff
@@ -0,0 +1,145 @@
...
+    } else {
+        println("No new annotations - nothing will be pushed.")
+        git.reset().setRef("HEAD~1").setMode(ResetCommand.ResetType.SOFT).call()
```
rgabbard: Comment on what this is doing in case it is not clear to a reader

annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
```diff
@@ -0,0 +1,145 @@
...
+    }
+}
+
+// Compare old data to newly exported data
+fun hasDiff(git: Git, newCommit: RevCommit): Boolean {
```
rgabbard: I think this can all be simplified to a call to http://download.eclipse.org/jgit/docs/jgit-2.0.0.201206130900-r/apidocs/org/eclipse/jgit/api/Status.html#isClean() ?

annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
```diff
@@ -0,0 +1,145 @@
...
+fun main(argv: Array<String>) {
+    val paramsLoader = SerifStyleParameterFileLoader.Builder()
+            .interpolateEnvironmentalVariables(true).build()
+    val params = paramsLoader.load(File(argv[0]))
+    val exportAnnotationsParams = params.getString("exportAnnotationsParams")
+    val restoreJsonParams = params.getString("restoreJsonParams")
+    val repoPath = params.getString("repoPath")
+
+    // Run ExportAnnotations.kt
+    ExportAnnotations.main(arrayOf(exportAnnotationsParams))
```
rgabbard: We should modify these parameters to put the correct directories in so the user doesn't need to coordinate it by hand. The easiest way to do this is to rename the current main methods of these to something which takes a `Parameters` object as its argument.  Then add `main` methods to them which loads the parameters and calls that method. That way we can call the parameters-as-args methods with modified parameters objects from this script.

annotation-utils/pom.xml
```diff
@@ -230,6 +230,13 @@
 			<version>0.12.0</version>
 		</dependency>
 
+		<!-- for managing git repositories -->
+		<dependency>
+			<groupId>org.eclipse.jgit</groupId>
+			<artifactId>org.eclipse.jgit</artifactId>
+			<version>5.4.0.201906121030-r</version>
```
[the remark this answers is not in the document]
elizlee: `version` removed

annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
elizlee: Moved `main` to top of file and used comment as docstring

annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
elizlee: Replaced `@code`s with backticks

annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
elizlee: For `exportAnnotationsParams` and `restoreJsonParams`, replaced `getString` with `getExistingFile`

annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
elizlee: Renamed function `pushNew` to `pushUpdatedAnnotations`

annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
elizlee: Removed `git.checkout()`

annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
elizlee: Considering adding comment: "If no changes are detected, remove the new commit"

annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
elizlee: Removed `pushIter` and `pushResult`

annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
elizlee: Used `logger.info` instead of `println`

annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
elizlee: Done

annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
elizlee: How did I not see this?
Done

rgabbard: JGit's documentation is not the greatest

annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
elizlee: Done

annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
elizlee: Done (for `repoToPushTo`)

annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
elizlee: Done

annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
elizlee: Sorry for the confusion - how does this prevent the user from needing to put the correct directory by hand?

rgabbard: So, as it is currently, the user needs to coordinate certain paths to "match up" between the three parameter files. We would rather generate those parameters automatically so that the user doesn't have to worry about it.

rgabbard: (and you never have to apologize for being confused. It's my job as a reviewer (and more senior developer) to be clear and your job to keep asking me questions until I manage to do so :-) )

elizlee: How would I automatically generate the parameters?

elizlee: So in `ExportAnnotations` for example, if I create a method out of `main` (call it `export`) and it takes a `Parameters` object as an argument, and then I call that method from `PushAnnotations`, where would the required `Parameters` come from if not specified by the user? I feel like I'm missing something here.

annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
elizlee: Done; wrote new method `setUpRepository` to perform actions depending on existence of `localWorkingCopyDirectory`

annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
elizlee: Done

annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
rgabbard: Almost! To make your comment the docstring for `main`, [you want to put it immediately before the function](https://kotlinlang.org/docs/reference/kotlin-doc.html) rather than just inside it (like Python doeS)

annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
```diff
@@ -0,0 +1,177 @@
...
+        git.add().addFilepattern("data/").call()
+        git.commit()
+                .setMessage("includes annotations from ${currentDate.getTime()}").call()
+        git.tag().setName(formatter.format(currentDate.getTime())).call()
+        pushLogger.info {"Pushing updated data"}
+        git.push()
+                .setTransportConfigCallback(SshConfig("~/.ssh/id_rsa"))
+                .call()
+    } else {
+        pushLogger.info {"No updates - nothing will be pushed"}
```
rgabbard: By convention, 99% of the time people call the logger for a scope just `logger` :-)

annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
```diff
@@ -0,0 +1,177 @@
...
+    }
+}
+
+fun setUpRepository(localWorkingCopyDirectory: String, repoToPushTo: String): Git {
```
rgabbard: When passing around files on the local working filesystem, whenever possible use [`File`](https://docs.oracle.com/javase/7/docs/api/java/io/File.html) or [`Path`](https://docs.oracle.com/javase/7/docs/api/java/nio/file/Path.html) rather than `String`.  Avoid being "stringly typed"! :-)

annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
```diff
@@ -0,0 +1,177 @@
...
+    /**
```
rgabbard: Like for `main`, shift this comment up above the method. A few other tips:

* by convention, the first line of the Javadoc for a method is a short, one-sentence description of what it does.  Then you have a blank line. Then you have a more detailed description. In this case, a good one-liner could be "Ensures [localWorkingCopy] is a checkout of [repoToPushTo] in a clean state"

annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
```diff
@@ -0,0 +1,177 @@
...
+     * Check if `localWorkingCopyDirectory` exists.
+     *   If it does: check the state
```
rgabbard: Say a bit more about what exactly it checks

annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
```diff
@@ -0,0 +1,177 @@
...
+     *   If not: clone `repoToPushTo` to the location specified by `localWorkingCopyDirectory`
+     */
+    if (File(localWorkingCopyDirectory).isDirectory()) {
+        val git = Git.open(File(localWorkingCopyDirectory))
+        val remoteUrl = git.getRepository()
```
rgabbard: Naming nitpick: calling this `currentRemoteUrl` or `workingCopyRemoteUrl` helps the reader differentiate it from the other remote URL floating around(`repoToPushTo`)

annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
```diff
@@ -0,0 +1,177 @@
...
+}
+
+fun setUpRepository(localWorkingCopyDirectory: String, repoToPushTo: String): Git {
+    /**
+     * Check if `localWorkingCopyDirectory` exists.
+     *   If it does: check the state
+     *   If not: clone `repoToPushTo` to the location specified by `localWorkingCopyDirectory`
+     */
+    if (File(localWorkingCopyDirectory).isDirectory()) {
+        val git = Git.open(File(localWorkingCopyDirectory))
```
rgabbard: If you take a look at [the Javadoc for `Git`](https://download.eclipse.org/jgit/site/5.4.0.201906121030-r/apidocs/index.html) you'll see the warning that "The caller is responsible for closing the repository; close() on this instance does not close the repo.". Presumably it has some cleanup it needs to do when you are done with it.

Kotlin and recent versions of Java make this easier by providing an interface called `AutoCloseable` for things which can be automatically closed (include on exit via exceptions) using special language constructs (`try`-with-resources for Java, `use` for Kotlin; Python has a similar thing called "context managers").  

In this case, you aren't done with the `Git` object in the local scope, so what you have in this method is fine, but back in `main` you want to change it to something like:
```
setUpRepository(localWorkingCopyDirectory, repoToPushTo).use { git ->
    // Run ExportAnnotations.kt
    pushLogger.info {"Beginning ExportAnnotations"}
    ExportAnnotations.export(exportAnnotationsParams)
    // Run RestoreJson.kt
    if (restoreJson) {
        pushLogger.info { "Beginning RestoreJson" }
        RestoreJson.restore(restoreJsonParams)
    }

    // Push new annotations
    pushUpdatedAnnotations(git)
    pushLogger.info {"Done!"}
}
```

At the close of the `use` block `.close()` will automatically be called on the `Git` object.

annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
```diff
@@ -0,0 +1,177 @@
...
+        val remoteUrl = git.getRepository()
+                .getConfig()
+                .getString("remote", "origin", "url")
+        // Does remote URL match `repoToPushTo`?
```
rgabbard: This comment and the next three are good examples of the sort of comments which can be omitted, especially given the text of the exception messages.  For example, _Is the status clean?_ doesn't really add anything to `if (!status.isClean())`.  In general, you want the comments to explain not the _what_ of the code but rather the _why_ of the code.

annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
```diff
@@ -0,0 +1,177 @@
...
+            throw RuntimeException("$localWorkingCopyDirectory has been modified locally.")
+        }
+        // Is it on the master branch?
+        val currentBranch = git.getRepository().getBranch()
+        if (currentBranch != "master") {
+            throw RuntimeException("$localWorkingCopyDirectory is not on the master branch.")
+        }
+        // Sync to remote master
+        git.fetch().setRemote("origin").call()
+        git.reset().setRef("origin/master").setMode(ResetCommand.ResetType.HARD).call()
```
rgabbard: In Git, (roughly speaking) your working copy stores its own copy of the state of the remote repository, but it only updates this copy when requested; you need to run a "fetch" here to make sure that what your working copy believes is `origin/master` truly is `origin/master`.

annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
rgabbard: I suggest putting this before `pushUpdatedAnnotations`; that way a user reading the file will encounter the methods in the order they will be executed.

annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
```diff
@@ -0,0 +1,177 @@
...
+        val currentDate = Calendar.getInstance(TimeZone.getTimeZone("UTC"))
+        TimeZone.setDefault(TimeZone.getTimeZone("UTC"))
+        val formatter = SimpleDateFormat("yyyy-MM-dd-HH-mm-ss")
+
+        git.add().addFilepattern("data/").call()
+        git.commit()
+                .setMessage("includes annotations from ${currentDate.getTime()}").call()
+        git.tag().setName(formatter.format(currentDate.getTime())).call()
+        pushLogger.info {"Pushing updated data"}
+        git.push()
```
rgabbard: Is there a way we can ensure the push was successful for we can alert the user by an exception if it wasn't?

annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
```diff
@@ -0,0 +1,177 @@
...
+
+    // A clean status indicates no updates
+    if (!status.isClean()) {
+        val currentDate = Calendar.getInstance(TimeZone.getTimeZone("UTC"))
+        TimeZone.setDefault(TimeZone.getTimeZone("UTC"))
+        val formatter = SimpleDateFormat("yyyy-MM-dd-HH-mm-ss")
+
+        git.add().addFilepattern("data/").call()
+        git.commit()
+                .setMessage("includes annotations from ${currentDate.getTime()}").call()
```
rgabbard: "through" is more accurate than "from" here

annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
rgabbard: Does this also push the tag?  If it doesn't you way need to also do `git.push().add(tag)....` where `tag` is the result of the call to `git.tag()....` above.

annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
```diff
@@ -0,0 +1,177 @@
...
+     * Postcondition: The exported annotations will be saved within a folder named `data/` in the local
+     * working directory.
+     */
+    if (argv.size != 1) {
+        throw RuntimeException("Expected a single argument: a parameter file")
+    }
+    val paramsLoader = SerifStyleParameterFileLoader.Builder()
+            .interpolateEnvironmentalVariables(true).build()
+    val params = paramsLoader.load(File(argv[0]))
+    val exportAnnotationsParamsBuilder = Parameters.builder()
```
rgabbard: `Parameters` has a method whose name I don't remember which will log the contents of the parameters. This can sometimes be useful for debugging, so I recommend it for most scripts.

annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
```diff
@@ -0,0 +1,177 @@
...
+    exportAnnotationsParamsBuilder.set("inceptionUrl", params.getString("inceptionUrl"))
+    exportAnnotationsParamsBuilder.set("inceptionUsername", params.getString("inceptionUsername"))
+    exportAnnotationsParamsBuilder.set("inceptionPassword", params.getString("inceptionPassword"))
+    exportAnnotationsParamsBuilder.set("exportedAnnotationRoot", params.getString("exportedAnnotationRoot"))
+    val exportAnnotationsParams = exportAnnotationsParamsBuilder.build()
+
+    val restoreJson = params.getString("restoreJson") == "yes"
+    val restoreJsonParamsBuilder = Parameters.builder()
+    if (restoreJson) {
```
rgabbard: A cleaner way to do this:

```kotlin
val restoreJsonParameters = if (params.getOptionalBoolean("restoreOriginalText").orElse(false)) {
   Parameters.builder()
        set("indexDirectory", params.getString("indexDirectory"))
       .set("gigawordDataDirectory", params.getString("gigawordDataDirectory"))
       .set("inputJsonDirectory", params.getString("inputJsonDirectory"))
       .set("restoredJsonDirectory", params.getString("restoredJsonDirectory"))
       .build()
} else {
   null
}
```

Then below you can do `if (restoreJsonParams != null) ...`

Also note in the snippet above suggested changes to the parameter name and default value.  Since it is now a boolean/ it will have values `true`/`false` rather than `yes`/`no`.

annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
```diff
@@ -0,0 +1,177 @@
...
+        throw RuntimeException("Expected a single argument: a parameter file")
+    }
+    val paramsLoader = SerifStyleParameterFileLoader.Builder()
+            .interpolateEnvironmentalVariables(true).build()
+    val params = paramsLoader.load(File(argv[0]))
+    val exportAnnotationsParamsBuilder = Parameters.builder()
+    exportAnnotationsParamsBuilder.set("inceptionUrl", params.getString("inceptionUrl"))
+    exportAnnotationsParamsBuilder.set("inceptionUsername", params.getString("inceptionUsername"))
+    exportAnnotationsParamsBuilder.set("inceptionPassword", params.getString("inceptionPassword"))
+    exportAnnotationsParamsBuilder.set("exportedAnnotationRoot", params.getString("exportedAnnotationRoot"))
```
rgabbard: The value for this parameter should be derived from `localWorkingCopyDirectory` so the user doesn't have to worry about it (and can't screw it up).

annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
```diff
@@ -0,0 +1,177 @@
...
+    val exportAnnotationsParams = exportAnnotationsParamsBuilder.build()
+
+    val restoreJson = params.getString("restoreJson") == "yes"
+    val restoreJsonParamsBuilder = Parameters.builder()
+    if (restoreJson) {
+        restoreJsonParamsBuilder.set("indexDirectory", params.getString("indexDirectory"))
+        restoreJsonParamsBuilder.set("gigawordDataDirectory", params.getString("gigawordDataDirectory"))
+        restoreJsonParamsBuilder.set("inputJsonDirectory", params.getString("inputJsonDirectory"))
```
rgabbard: As above, both `inputJsonDirectory`  and `restoredJsonDirectory` should be derived from `localWorkingCopyDirectory`

annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
elizlee: Ah, for some reason I thought I couldn't use `logger` from `ExportAnnotations`

annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
elizlee: Done - same for `setUpRepository`

annotation-utils/src/main/java/edu/isi/vista/annotationutils/PushAnnotations.kt
elizlee: Changed `localWorkingCopyDirectory` to a `File` object

PR #38, opened 2019-09-05
modifies pom.xml

elizlee:


pom.xml
```diff
@@ -161,6 +161,13 @@
 					<groupId>org.codehaus.mojo</groupId>
 					<artifactId>appassembler-maven-plugin</artifactId>
 					<version>${appassembler.maven.plugin.version}</version>
+					<configuration>
+						<programs>
+							<program>
+								<mainClass>edu.isi.vista.annotationutils.PushAnnotations</mainClass>
+							</program>
+						</programs>
```
berquist: This block should go in https://github.com/isi-vista/curated-training-annotator/blob/master/annotation-utils/pom.xml.

No AppAssembler script is being made because you need a `<id>myCommandLineScriptName</id>` entry at the same level as the `mainClass` field.

berquist:
Actually, I don't have merge access, so you probably don't either. @rgabbard can you take a look at the repo settings?

PR #40, opened 2019-09-26
Add script for extracting statistics from JSON exports

elizlee:


annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExtractAnnotationStats.kt
```diff
@@ -0,0 +1,223 @@
...
+            TimeZone.setDefault(TimeZone.getTimeZone("UTC"))
+            val formatter = SimpleDateFormat("yyyy-MM-dd")
+            val thisDate = formatter.format(currentDate.getTime())
+
+            val annotationSheet: MutableList<Annotation> = arrayListOf()
+
+            // Go through each file in each user's folder and collect the stats
+            var user: String? = null
+            var subtype: String? = null
+            File(exportAnnotationRoot.toString()).walk().forEach {
```
rgabbard: Isn't `exportedAnnotationRoot` already a `File`?

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExtractAnnotationStats.kt
```diff
@@ -0,0 +1,223 @@
...
+                // Skip irrelevant dirs
+                if (it == File(exportAnnotationRoot.toString()) ||
```
rgabbard: nit: A more Kotlin-esque way to do this is to do 

```kotlin
exportAnnotationRoot.walk()
   .filterNot { it.name.contains("sandbox") || it.name.contains("gigaword") }
   .forEach { 
    // stuff
   }
```

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExtractAnnotationStats.kt
```diff
@@ -0,0 +1,223 @@
...
+                        it.name.contains("""sandbox""") ||
+                        it.name.contains("""gigaword""")) {
+                }
+                else if (it.isDirectory) {
+                    val folder = it.name
+                    // Collect username, subtype, and subSubtype
+                    user = Regex(pattern = """\.\w*-(\w*)""")
```
rgabbard: Can you leave a comment clarifying these regexes? (e.g. with sample names?)

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExtractAnnotationStats.kt
```diff
@@ -0,0 +1,223 @@
...
+                            .find(input = folder.toString())!!.groupValues[1]
```
rgabbard: Avoid using `toString` for program logic - I think what you want here is `folder.absolutePath`.

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExtractAnnotationStats.kt
```diff
@@ -0,0 +1,223 @@
...
+            // Go through each file in each user's folder and collect the stats
+            var user: String? = null
+            var subtype: String? = null
+            File(exportAnnotationRoot.toString()).walk().forEach {
+                // Skip irrelevant dirs
+                if (it == File(exportAnnotationRoot.toString()) ||
+                        it.name.contains("""sandbox""") ||
+                        it.name.contains("""gigaword""")) {
+                }
+                else if (it.isDirectory) {
```
rgabbard: nit: I think it would be clearer to ignore directories and extract the user and event type by examining the parent directories of the files you find

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExtractAnnotationStats.kt
```diff
@@ -0,0 +1,223 @@
...
+                    val folder = it.name
+                    // Collect username, subtype, and subSubtype
+                    user = Regex(pattern = """\.\w*-(\w*)""")
+                            .find(input = folder.toString())!!.groupValues[1]
+                    subtype = Regex(pattern = """(\w*\.\w*)-""")
```
rgabbard: nit: what is the type "sub" of? Would plain `eventType` be clearer?

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExtractAnnotationStats.kt
```diff
@@ -0,0 +1,223 @@
...
+                            .find(input = folder.toString())!!.groupValues[1]
+                }
+                else if (it.isFile) {
+                    if (user.isNullOrBlank() || subtype.isNullOrBlank()) {
+                        throw RuntimeException("Missing annotation information (user/subtype)")
```
rgabbard: Include the path to the file in the error message to make debugging easier.

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExtractAnnotationStats.kt
```diff
@@ -0,0 +1,223 @@
...
+                    }
+                    val documentName: String = it.toString()
```
rgabbard: `it.name`

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExtractAnnotationStats.kt
```diff
@@ -0,0 +1,223 @@
...
+                    // TODO: count events rather than annotated documents
+                    // This is sort of a placeholder for a better method
```
rgabbard: make an issue for this and reference it here

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExtractAnnotationStats.kt
```diff
@@ -0,0 +1,223 @@
...
+                    val jsonString: String = File(it.toString()).readText(Charsets.UTF_8)
```
rgabbard: `it` is already a file, right?

rgabbard: nit: you don't need to supply the encoding if it's UTF-8 because that's the default: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io/java.io.-file/read-text.html

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExtractAnnotationStats.kt
```diff
@@ -0,0 +1,223 @@
...
+                    if (jsonString.contains(Regex(""""CTEventSpan" : \["""))) {
```
rgabbard: PRefer to factor out regular expressions as constants so they don't need to get repeatedly compiled.

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExtractAnnotationStats.kt
```diff
@@ -0,0 +1,223 @@
...
+                        annotationSheet.add(Annotation(documentName, user!!, subtype!!))
+                    }
+                }
+            }
+            // From annotationSheet, get TOTAL, TOTAL PER USER/SUBTYPE/SUBSUBTYPE
+            val totalAnnotations = annotationSheet.size
```
rgabbard: I am confused by the `Sheet` in the name `annotationSheet`. Perhaps clearer would be `documentsWithEventAnnotations`?

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExtractAnnotationStats.kt
```diff
@@ -0,0 +1,223 @@
...
+                if (cols.isNotEmpty()) {
+                    subtypeMap[cols[0].text()] = cols[1].text().toInt()
+                }
+            }
+            return previousAnnotations
+        }
+    }
+}
+
+data class Annotation(val documentName: String, val user: String, val subtype: String)
```
rgabbard: 💯 for pulling out a class for the annotated documents.  Since currently we are only tracking things at the document level, how about calling it `DocumentAnnotation`?

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExtractAnnotationStats.kt
```diff
@@ -0,0 +1,223 @@
...
+        }
+        fun extractStats(params: edu.isi.nlp.parameters.Parameters) {
+            val exportAnnotationRoot = params.getExistingDirectory("exportedAnnotationRoot")
+            val statisticsDirectory = params.getExistingDirectory("statisticsDirectory")
+            val currentDate = Calendar.getInstance(TimeZone.getTimeZone("UTC"))
+            TimeZone.setDefault(TimeZone.getTimeZone("UTC"))
+            val formatter = SimpleDateFormat("yyyy-MM-dd")
+            val thisDate = formatter.format(currentDate.getTime())
+
+            val annotationSheet: MutableList<Annotation> = arrayListOf()
```
rgabbard: nit: Consider pulling the entire process of building `annotationSheet` into its own method to make it easier for the user to follow the flow of `extractStats`

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExtractAnnotationStats.kt
```diff
@@ -0,0 +1,223 @@
...
+                    // This is sort of a placeholder for a better method
+                    val jsonString: String = File(it.toString()).readText(Charsets.UTF_8)
+                    if (jsonString.contains(Regex(""""CTEventSpan" : \["""))) {
+                        annotationSheet.add(Annotation(documentName, user!!, subtype!!))
+                    }
+                }
+            }
+            // From annotationSheet, get TOTAL, TOTAL PER USER/SUBTYPE/SUBSUBTYPE
+            val totalAnnotations = annotationSheet.size
+            val annotationsPerUser = annotationSheet.countBy { it.user }
```
rgabbard: would `annotationsByUser` be clearer?

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExtractAnnotationStats.kt
```diff
@@ -0,0 +1,223 @@
...
+            val formatter = SimpleDateFormat("yyyy-MM-dd")
+            for (day in 1..7) {
+                calendar.add(Calendar.DAY_OF_YEAR, -1)
+                val lastDate = formatter.format(calendar.getTime())
+                val lastFilename = "$statsDir/StatsReport$lastDate.html"
+                if (File(lastFilename).exists()) {
+                    return lastFilename
+                }
+            }
+            return ""
```
rgabbard: Avoid using the empty string to represent things which are missing. Prefer to return `Optional<String>` so that the type system can make sure the user takes into account the possibility that the returned value might be missing.

rgabbard: You might actually want to return a data class object here so you could also bundle in the date. That way your HTML can tell how indicate when the snapshot it is comparing against was.

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExtractAnnotationStats.kt
```diff
@@ -0,0 +1,223 @@
...
+                }
+            }
+            }
+        }
+        /**
+         * Search as far back as a week for a previous stats report
+         * The length of time can be changed
+         */
+        fun locatePreviousStats(currentDate: Calendar, statsDir: File): String {
+            val calendar = currentDate
```
rgabbard: nit: This works, but a more elegant solution would be to take all the matching files in the directory, extract dates from them, sort them, and take the latest.

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExtractAnnotationStats.kt
```diff
@@ -0,0 +1,223 @@
...
+                    }
+                }
+            }
+            }
+        }
+        /**
+         * Search as far back as a week for a previous stats report
+         * The length of time can be changed
+         */
+        fun locatePreviousStats(currentDate: Calendar, statsDir: File): String {
```
rgabbard: Avoid using the `String` data type to represent things which aren't strings. In this case, return `File`.

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExtractAnnotationStats.kt
```diff
@@ -0,0 +1,223 @@
...
+
+            val lastFilename = locatePreviousStats(currentDate, statisticsDirectory)
+            val previousAnnotationsPerUser: MutableMap<String, Int> = mutableMapOf()
+            val previousAnnotationsPerSubtype: MutableMap<String, Int> = mutableMapOf()
+            var previousAnnotations = 0
+            if (lastFilename.isEmpty()) {
+                logger.info {"No previous stats report in over a week"}
+            }
+            else {
+                previousAnnotations = getPreviousAnnotations(File(lastFilename), previousAnnotationsPerUser, previousAnnotationsPerSubtype)
```
rgabbard: There is no need to declare  `previousAnnotationsPerUser` and `previousAnnotationsPerSubType` here and then pass them in to `getPreviousAnnotation`; instead, `getPreviousAnnotation` can build them directly and return the pair via a data class object.

nit: my naming preference is to prefer `load` over `get` as a prefix for something which touches disk or DBs or other non-memory resources.

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExtractAnnotationStats.kt
```diff
@@ -0,0 +1,223 @@
...
+            }
+            // From annotationSheet, get TOTAL, TOTAL PER USER/SUBTYPE/SUBSUBTYPE
+            val totalAnnotations = annotationSheet.size
+            val annotationsPerUser = annotationSheet.countBy { it.user }
+            val annotationsPerSubtype = annotationSheet.countBy { it.subtype }
+            // For better organization
+            val sortedUsers = annotationsPerUser.toSortedMap()
+            val sortedSubtypes = annotationsPerSubtype.toSortedMap()
+
+            val lastFilename = locatePreviousStats(currentDate, statisticsDirectory)
```
rgabbard: All the logic for loading the previous annotations should be pulled out into its own method

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExtractAnnotationStats.kt
```diff
@@ -0,0 +1,223 @@
...
+            var previousAnnotations = 0
+            if (lastFilename.isEmpty()) {
+                logger.info {"No previous stats report in over a week"}
+            }
+            else {
+                previousAnnotations = getPreviousAnnotations(File(lastFilename), previousAnnotationsPerUser, previousAnnotationsPerSubtype)
+            }
+
+            // Convert values to html
+            File("$statisticsDirectory/StatsReport$thisDate.html").printWriter().use{
```
rgabbard: prefer
```
File(statisticsDirectory, "StatsReport$thisDate.html")
```

(again, avoid working with strings instead of using the provided logic)

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExtractAnnotationStats.kt
```diff
@@ -0,0 +1,223 @@
...
+                calendar.add(Calendar.DAY_OF_YEAR, -1)
+                val lastDate = formatter.format(calendar.getTime())
+                val lastFilename = "$statsDir/StatsReport$lastDate.html"
+                if (File(lastFilename).exists()) {
+                    return lastFilename
+                }
+            }
+            return ""
+        }
+        fun getPreviousAnnotations(
```
rgabbard: Rather than parsing this out of the HTML, how about saving the results each time as JSON and loading the previous stats from there?

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExtractAnnotationStats.kt
elizlee: Whoops - changed

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExtractAnnotationStats.kt
elizlee: Wasn't aware of this method - thanks!

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExtractAnnotationStats.kt
elizlee: I was originally using terminology from "Instructions for Annotators"

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExtractAnnotationStats.kt
elizlee: Just `folder` also seems to work

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExtractAnnotationStats.kt
elizlee: ...Yeah

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExtractAnnotationStats.kt
elizlee: Sounds good

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExtractAnnotationStats.kt
```diff
@@ -0,0 +1,283 @@
...
+        fun extractStats(params: edu.isi.nlp.parameters.Parameters) {
+            val exportAnnotationRoot = params.getExistingDirectory("exportedAnnotationRoot")
+            val statisticsDirectory = params.getExistingDirectory("statisticsDirectory")
+            val currentDate = Calendar.getInstance(TimeZone.getTimeZone("UTC"))
+            TimeZone.setDefault(TimeZone.getTimeZone("UTC"))
+            val formatter = SimpleDateFormat("yyyy-MM-dd")
+            val thisDate = formatter.format(currentDate.getTime())
+
+            val documentsWithEventAnnotations: MutableList<DocumentAnnotation> = arrayListOf()
+            collectStats(exportAnnotationRoot, documentsWithEventAnnotations)
```
rgabbard: Except in a few special cases (e.g. internal methods of graph traversals), avoid the pattern of "Create a collection and pass it as an argument of a method which will populate it". Instead, drop the argument and simple return the data:

```kotlin
val documentsWithEventAnnotations = collectStates(exportAnnotationRoot)
```

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExtractAnnotationStats.kt
```diff
@@ -0,0 +1,283 @@
...
+            // For better organization
+            val sortedUsers = annotationsByUser.toSortedMap()
+            val sortedEventTypes = annotationsByEventType.toSortedMap()
+            val newAnnotationStats = AnnotationStats(totalAnnotations, sortedUsers, sortedEventTypes)
+
+            // Load the annotation statistics from the last run
+            val previousStatsReport: StatsReport? = locatePreviousStats(statisticsDirectory)
+            var previousAnnotationStats: AnnotationStats? = null
+            var annotationDiffs: AnnotationStats? = null
+            if (previousStatsReport == null) {
```
rgabbard: nit: it's somewhat more readable if you put the "expected/more interesting" branch of the "if" first and the warning/error condition last.

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExtractAnnotationStats.kt
```diff
@@ -0,0 +1,283 @@
...
+                            }
+                        }
+                    }
+        }
+
+        /**
+         * Find all existing JSON files in the output directory and return
+         * the most recent file; else return null
+         */
+        fun locatePreviousStats(statsDir: File): StatsReport? {
```
rgabbard: nit: This is a fine way to write it. Another Kotlin-y way might be:

```
return statsDir.walk().asSequence()
        .filter { it.name.endsWith(".json") } }
        .map {  val datePattern = Regex(pattern = """\d{4}-\d{2}-\d{2}""")
                    val fileDate = datePattern
                            .find(input = it.name)?.value
                    if (fileDate != null) {
                        StatsReport(it, fileDate)
                    } else {
                        logger.debug("No date found for JSON file ${it.name}")
                        null
                    } 
               }
         .filterNotNull()
         .sortedBy { it.report.lastModified() }
         .lastOrNull()
```

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExtractAnnotationStats.kt
```diff
@@ -0,0 +1,283 @@
...
+        }
+
+        /**
+         * Use the numbers in the new statistics and previous ones to get the differences
+         */
+        fun getDiffs(newStats: AnnotationStats, previousStats: AnnotationStats): AnnotationStats {
+            val totalDiff = newStats.total - previousStats.total
+            val userDiff: MutableMap<String, Int?> = mutableMapOf()
+            for (user in newStats.byUser.keys) {
+                if (previousStats.byUser[user] != null)
```
rgabbard: Always use `{`s on your ifs: https://embeddedgurus.com/barr-code/2014/03/apples-gotofail-ssl-security-bug-was-easily-preventable/

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExtractAnnotationStats.kt
```diff
@@ -0,0 +1,283 @@
...
+            val mapper = jacksonObjectMapper()
+            return mapper.readValue<AnnotationStats>(statsFile)
+        }
+
+        /**
+         * Use the numbers in the new statistics and previous ones to get the differences
+         */
+        fun getDiffs(newStats: AnnotationStats, previousStats: AnnotationStats): AnnotationStats {
+            val totalDiff = newStats.total - previousStats.total
+            val userDiff: MutableMap<String, Int?> = mutableMapOf()
```
rgabbard: nit: Another option:

```kotlin
val userDiff = newStats.filterKeys { oldStats.containsKey(it) }
                       .mapValues { (user, newCount) -> newCount - oldStats[it] }
```

rgabbard: (this also lets you make the value type `Int` rather than `Int?`

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExtractAnnotationStats.kt
elizlee: Oops, that was a mistake

PR #54, opened 2020-02-03
Add step to ElasticSearch setup instructions

elizlee:


docs/SETTING_UP_ELASTICSEARCH.md
```diff
@@ -8,21 +8,30 @@ tar xvzf gigaword_eng_5_LDC2011T07.tgz
...
-```
+    ```
+    tar xvzf elasticsearch-6.6.1.tar.gz
+    ```
 
-3. Check service status:
-``` 
-curl -X GET http://localhost:9200
-```
+    On CentOS, the following content needs to be in the file /etc/sysctl.d/90-elastic-search.conf:
```
gabbard:
```suggestion
    On CentOS, the following content needs to be in the file `/etc/sysctl.d/90-elastic-search.conf`:
```

gabbard: While you are at it, can you add similar single-backticks around any paths in the file which are missing them (e.g. in step 1).

PR #56, opened 2020-02-24
Ace Indexer Module added

### LINKED ISSUE #55: Load ACE corpus into Inception for clue word annotation
We would like to load the ACE corpus into Inception so we can annotate clue words on its events.

This requires a few steps:
1. get the original text of the training section of the ACE documents.
2. figure out how to create an Inception project containing those documents and the existing ACE event annotation, translated into our curated training annotation format.

aakum03:
Closes #55


ace-indexer/src/main/java/edu/isi/vista/aceIndexer/IndexAceWithElasticSearch.java
```diff
@@ -0,0 +1,262 @@
...
+
+    private static Article articleFromPath(Path filePath) throws IOException{
+
+        String docString = new String(Files.readAllBytes(filePath));
+        // DOC ID marker
+        Pattern ACE_DOC_ID_PATTERN = Pattern.compile("<DOCID> (.*?) <.DOCID>");
+
+        Matcher m =
+                ACE_DOC_ID_PATTERN.matcher(
+                        docString.substring(0, Math.min(120, docString.length())));
```
gabbard: Why the `Math.min`?

ace-indexer/pom.xml
```diff
@@ -0,0 +1,125 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>curated-training-all</artifactId>
+        <groupId>edu.isi.vista</groupId>
+        <version>0.1.0-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>ace-indexer</artifactId>
+    <dependencies>
```
elizlee: Recommendation: use the `appassembler` plugin to create an executable script. This way, after building the project, one can easily run the indexer through the command line with something like
`ace-indexer/target/appassembler/bin/indexAce path/to/file.params`.

See https://github.com/isi-vista/curated-training-annotator/blob/master/gigaword-indexer/pom.xml for reference.

ace-indexer/src/main/java/edu/isi/vista/aceIndexer/IndexAceWithElasticSearch.java
```diff
@@ -0,0 +1,262 @@
...
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/**
+ * Indexes Ace with Elastic Search in a way usable by the external search feature of the Inception
+ * annotator.
+ */
+public class IndexAceWithElasticSearch {
+    //TODO: Modify these completely:
```
elizlee: Do these still need modifying? How so?

ace-indexer/src/main/java/edu/isi/vista/aceIndexer/IndexAceWithElasticSearch.java
```diff
@@ -0,0 +1,262 @@
...
+
+    private static final int DEFAULT_PORT_PRI = 9200;
+
+    private static final int DEFAULT_PORT_SEC = 9201;
+
+    private static final String PARAM_CORPUS_DIRECTORY_PATH = "corpusDirectoryPath";
+
+    private static final String PARAM_FORMAT = "format";
+
+    // Still Needed? since the ACE corpus comes as a single .tgz containing all the documents
```
elizlee: This parameter could still be useful to someone who e.g. had downloaded the ACE corpus and `gunzip`-ed it

ace-indexer/src/main/java/edu/isi/vista/aceIndexer/IndexAceWithElasticSearch.java
aakum03: No, I will be removing this comment

ace-indexer/src/main/java/edu/isi/vista/aceIndexer/IndexAceWithElasticSearch.java
aakum03: This is also in the existing gigaword indexer. I believe it is for the edge case where the entire document is less that 120 characters long. To prevent a `StringIndexOutOfBoundsException` from `substring()`

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/IndexGigawordWithElasticSearch.java
```diff
@@ -121,40 +125,74 @@ public static void main(String[] argv) throws IOException {
...
-                                "Indexing terminated early without error, probably due to the user "
-                                        + "requesting a limit on the number of documents indexed");
-                        return false;
-                      }
-                  } catch (Exception e) {
-                    throw new RuntimeException(e);
-                  }
-                  return true;
-                });
+      if (!format.equalsIgnoreCase("ace")) {
```
gabbard: Is it possible to refactor so that more of the code is shared between the ACE and not-ACE cases? There seems to be a lot of duplication.

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/AceDocument.java
```diff
@@ -0,0 +1,52 @@
...
+    @Override
+    public void close() throws Exception
+    {
+
+    }
+
+    private class ArticleIterator extends AbstractIterator<Article> {
+        protected Article computeNext(){
+            // DOC ID marker
+            Pattern ACE_DOC_ID_PATTERN = Pattern.compile("<DOCID> (.*?) <.DOCID>");
```
gabbard: Why not `/DOCID`?

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/IndexGigawordWithElasticSearch.java
```diff
@@ -121,13 +125,29 @@ public static void main(String[] argv) throws IOException {
       PathMatcher filePattern;
       if (format.equalsIgnoreCase("LTF")) {
         filePattern = FileSystems.getDefault().getPathMatcher("glob:**.ltf.zip");
+      } else if (format.equalsIgnoreCase("ace")){
+        // Source files (unannotated) are in the .sgm format
+        // Only checks each adj subdirectory to avoid duplicate source files
+        // (as fp1, fp2 and timex2 use the same source files)
+        if (lang.equalsIgnoreCase("english")) {
```
gabbard: nit: more elegant is:

```java
private static final languageToAceDirectoryName = ImmutableMap.of(
    "english", "English",
    "chinese", "Chinese".
    "arabic", "Arabic");

...


final Matcher filePattern;

...

final String aceDirectoryName = languageToAceDirectoryName.get(lang.toLowerCase());

if (aceDirectoryName ==null) {
	// exception here
}

filePattern = FileSystems.getDefault().getPathMatcher("glob:**/" + aceDirectoryName + "/**/adj/*.sgm")
```

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/AceDocument.java
```diff
@@ -0,0 +1,52 @@
...
+
+            Matcher m =
+                    ACE_DOC_ID_PATTERN.matcher(
+                            docText.substring(0, Math.min(120, docText.length())));
+            if (m.find()) {
+                String docId = m.group(1);
+                return new Article(docId, docText);
```
gabbard: This is going to feed all the markup directly into ElasticSearch, which will be confusing for the annotators and might mess up the HTML display. We need to clean it in a way which preserves offsets.

Is there documentation in the ACE corpus about how offsets are computed?

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/AceDocument.java
aakum03: I was unsure if an escape character was needed for '/' since I was seeing conflicting information (about needing an escape character). I've since found that the escape character is not needed. I will make the change.

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/IndexGigawordWithElasticSearch.java
aakum03: Change has been made

gigaword-indexer/src/main/java/edu/isi/vista/gigawordIndexer/AceDocument.java
aakum03: I've been able to find that the XML tags are not counted and end-of-lines are counted as a single character. How should I proceed with the cleanup?

gabbard: Transform the original text which gets indexed by elastic search to remove all content in HTML tags and to collapse any Windows CRLF EOLs to `\n`.

aakum03: Added. I did this by removing all tags <.*?>.  however, one thing affected would be the source mentioned inside the tag. E.g: 
<DOCTYPE SOURCE="broadcast news"> NEWS STORY </DOCTYPE>
transforming into
NEWS STORY
Is this acceptable? (This appears uniformly mapped. I.e: all "broadcast news" have NEWS STORY)

gabbard: That is not a problem.

gabbard: Actually can you experiment with importing a few documents with the markup left in-place (but the newlines still fixed) into Inception and see how it appears in the interface? @elizlee can probably help you with this. If it isn't garbled, it may be best to leave the markup in place.

PR #60, opened 2020-03-24
Apf ingester to generate ACE corpus projects

aakum03:
This is what I currently have on the apf ingester. I am waiting on the next Inception release


aakum03:
This is the nearly complete version. Testing worked as expected, importing desired annotated documents into a single project.
I still need to:
- add support for a parameters file
- make sure there arent any windows/unix path type conflicts
- write a usage instructions
Note: cached files are not in this draft (those directories were added to the git ignore)

annotation-utils/src/main/java/edu/isi/vista/annotationutils/RestoreJson.kt
```diff
@@ -1,109 +1,149 @@
...
+private fun getAceDocID(filename: String): String {
+    // Returns empty string if not an AceFileName
+    // Ace filenames contain an event.subtype in the file name
+    // Example: CNNHL_ENG_20030304_142751.10-Business.Declare-Bankruptcy.xmi-liz_lee.json
+    val regex = """(.*_[^a-zA-Z]*)-(.*?)-.*?""".toRegex()
+    if (regex.containsMatchIn(filename)) {
+        val matchResult = regex.find(filename)
+        return matchResult!!.groups[1]!!.value
+    } else {
+        return ""
```
elizlee, line 147: Return `null` instead of an empty string. Also, since you have a return statement in the first block, you don't need to put `return null` in an `else` block.

annotation-utils/src/main/java/edu/isi/vista/annotationutils/RestoreJson.kt
```diff
@@ -1,109 +1,149 @@
...
+                            jsonFile.toPath().removePrefixPath(inputJsonDirectory.toPath()).toString()
+                    ).parent.toAbsolutePath()
+                    if (Files.notExists(projectOutDir)) {
+                        Files.createDirectories(projectOutDir)
+                    }
+                    val jsonTree = objectMapper.readTree(jsonFile)
+                    var text = ""
+                    // If aceDocID is empty, it means the document being processed is not an
+                    // ace document
+                    if(aceDocID.isNotEmpty()) {
```
elizlee, line 72: Instead of initializing `text` before this if-statement, you can do something like this:
```
val text = if (aceDocID.isNotEmpty()) {
  ...
  aceTextSource.getOriginalText(docID).orNull()
    ?: throw RuntimeException("Could not get original text for $docID")
} else {
  ...
```

annotation-utils/annotation-utils.iml
```diff
@@ -0,0 +1,165 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
gabbard, line 1: Don't check in IntelliJ IDEA `.iml` files - instead people cloning the repo and wishing to use IntelliJ IDEA should import the root `pom.xml`.

annotation-utils/pom.xml
```diff
@@ -1,333 +1,413 @@
...
-	<repositories>
-		<repository>
-			<id>jcenter</id>
-			<name>jcenter</name>
-			<url>http://jcenter.bintray.com</url>
-		</repository>
-	</repositories>
-
-</project>
+<?xml version="1.0" encoding="UTF-8"?>
```
gabbard, line 1: Were there any substantive changes here or is this just reformatting?

annotation-utils/src/main/java/edu/isi/vista/annotationutils/AceCorpusTextSource.java
```diff
@@ -0,0 +1,52 @@
...
+import java.nio.file.Path;
+import java.io.IOException;
+
+
+public final class AceCorpusTextSource implements OriginalTextSource{
+    private  File aceDataDirectory;
+    private File[] sourceFileDirectories;
+
+    public Optional<String> getOriginalText(final Symbol docID) throws IOException {
+        String rawSourceText = "";
```
gabbard, line 18: Avoid using empty strings to represent anything except an actual empty string.  Here I think you are using it to represent "we couldn't find the source raw text", so you want the type of `rawSourceText` to be `Optional<String>`, initialized to `Optional.absent()`.  Then set it to `Optional.of(....)` if you find it.

annotation-utils/src/main/java/edu/isi/vista/annotationutils/AceCorpusTextSource.java
```diff
@@ -0,0 +1,52 @@
...
+import com.google.common.base.Optional;
+import edu.isi.nlp.io.OriginalTextSource;
+import java.io.File;
+import java.nio.file.Path;
+import java.io.IOException;
+
+
+public final class AceCorpusTextSource implements OriginalTextSource{
+    private  File aceDataDirectory;
+    private File[] sourceFileDirectories;
```
gabbard, line 15: style: Unless you are writing low-level code, you almost never want to use arrays in Java. Prefer `List<File>` (or `List<Path>`).

annotation-utils/src/main/java/edu/isi/vista/annotationutils/AceCorpusTextSource.java
```diff
@@ -0,0 +1,52 @@
...
+                break;
+            }
+        }
+        if(rawSourceText.isEmpty()) {
+            throw new IOException("Source file is empty for document: " + docID.toString());
+        }
+        return Optional.of(processAceSourceText(rawSourceText));
+    }
+
+    AceCorpusTextSource(File aceDataDirectory) throws IOException{
```
gabbard, line 33: style: You don't want your constructors to do any work. Prefer writing a private constructor which simply assigns its arguments to the fields. Then shift the logic from here into a `public static AceCorpusTextSource fromDirectory(....)` factory function.

Homework assignment: read, mark, learn, and inwardly digest Joshua Bloch's _Effective Java_. :-)

annotation-utils/src/main/java/edu/isi/vista/annotationutils/AceCorpusTextSource.java
```diff
@@ -0,0 +1,52 @@
...
+        this.sourceFileDirectories = aceDataDirectory.listFiles();
+        for(int i = 0; i < this.sourceFileDirectories.length; i++) {
+            this.sourceFileDirectories[i] = new File(sourceFileDirectories[i], "adj");
+        }
+    }
+
+    private String processAceSourceText (String rawSourceText) {
+        return rawSourceText
+                .replaceAll("\\r\\n", "\n")
+                .replaceAll("<[\\s\\S]*?>", "");
```
gabbard, line 50: Why not `.*?`?

gabbard: Add a comment explaining to the reader why these replacements are made (ACE offset counting normalizes line endings and does not include markup).

annotation-utils/src/main/java/edu/isi/vista/annotationutils/RestoreJson.kt
```diff
@@ -1,109 +1,148 @@
-package edu.isi.vista.annotationutils
```
gabbard, line 1: What happened to make this giant diff for this file? It makes it hard to review for changes.

apf-ingester/apf_Ingester.py
```diff
@@ -0,0 +1,264 @@
+import re
```
gabbard, line 1: Name should be `apf_ingester.py` - caps are not used in Python module names

apf-ingester/apf_Ingester.py
```diff
@@ -0,0 +1,264 @@
...
+import json
+import shutil
+import tempfile
+import pathlib
+from vistautils.parameters import Parameters
+from vistautils.parameters_only_entrypoint import parameters_only_entry_point
+from typing import Dict, List, Optional
+
+
+def cleanup_source_document(sgm_path: str) -> None:
```
gabbard, line 12: style: Don't use strings for paths - use `pathlib.Path`.

apf-ingester/apf_Ingester.py
```diff
@@ -0,0 +1,264 @@
...
+    """Removes CRLF EOLs and in-text markup  in the file to preserve string offsets used for
+       annotation (as the ACE annotation offsets use '\n' EOLs and ignore markup)
+
+    Args:
+        sgm_path (str): String path of the .sgm file
+
+    One of sgm_string or sgm_path must be specified
```
gabbard, line 19: I don't see an `sgm_string` argument?

apf-ingester/apf_Ingester.py
```diff
@@ -0,0 +1,264 @@
...
+
+    Returns:
+        None
+
+    """
+    with open(sgm_path, 'r') as sgm_file:
```
gabbard, line 25: If you make `sgm_path` a `Path`, you can just do `sgm_string = sgm_path.read_text(encoding="utf-8")`

apf-ingester/apf_Ingester.py
gabbard, line 12: In-place changes should be avoided - prefer to write to a new location and return the path to it, or if possible just work with strings in-memory.

apf-ingester/apf_Ingester.py
```diff
@@ -0,0 +1,264 @@
...
+        visited_keys = []
+        for event in re.findall(r"<event ID=.*?>", self.apf_string):
+            matcher = re.fullmatch(r'<event ID="(.*?)-EV(.*?)" TYPE="(.*?)" SUBTYPE="(.*?)".*?>',
+                                   event)
+            if matcher is not None:
+                key_name: str = matcher.group(3) + "." + matcher.group(4)
+                doc_id: str = matcher.group(1)
+                if key_name in visited_keys:
+                    continue
+                if key_name in current_project_dict:
```
gabbard, line 85: style: Use a `defaultdict(list)` and you can avoid the `if/else` here.

apf-ingester/apf_Ingester.py
```diff
@@ -0,0 +1,264 @@
...
+            source_path = ace_corpus_path + filename
+            apf_path = ace_corpus_path + filename_without_extension + ".apf.xml"
+            ingester_input = ApfIngesterInput(apf_path)
+            ingester_input.extract_event_list_to_dict(complete_project_map)
+    # Add empty entry for 'None' type, to generate an empty project
+    complete_project_map["None"] = []
+    return complete_project_map
+
+
+def configure_and_generate_project(json_template_path: str,
```
gabbard, line 117: When you have a lot of arguments of similar type, prefer to make them keyword only (in this case, place a `*` first in the argument list, like `(*, `.

apf-ingester/apf_Ingester.py
```diff
@@ -0,0 +1,264 @@
...
+    if not os.path.exists(cached_ace_data_path):
+        os.makedirs(cached_ace_data_path)
+
+    # List of users (strings)
+    user_list = params.arbitrary_list("user_list")
+    # List of event type (Format: "EVENT_TYPE.SUBTYPE" strings)
+    event_list = params.arbitrary_list("event_list")
+
+    # Output Directory Path where configured projects are moved to (use an empty directory)
+    output_dir_path = params.string("output_dir_path")
```
gabbard, line 235: `params.creatableDirectory()`

apf-ingester/apf_Ingester.py
```diff
@@ -0,0 +1,264 @@
...
+        for filename in os.listdir(corpus_path):
+            if filename.endswith(".sgm") or filename.endswith(".apf.xml"):
+                shutil.copyfile(corpus_path + filename, destination_path + filename)
+
+
+def main(params: Parameters):
+    # List of the six ACE corpus /adj/ folders (one for each type: bc, bn, cts, nw, un, wl)
+    corpus_paths = params.arbitrary_list("corpus_paths")
+    # Path to the project config file template (json file)
+    json_template_path = params.string("json_template_path")
```
gabbard, line 215: `params.existing_directory()`

annotation-utils/src/main/java/edu/isi/vista/annotationutils/AceCorpusTextSource.java
aakum03: Some of the tags (in rare cases) had its opening and closing symbols on different lines which `.*?` does not include, so I decided to use `<[\\s\\S]*?>`

annotation-utils/src/main/java/edu/isi/vista/annotationutils/RestoreJson.kt
aakum03: I'm not sure why this happened. Is there any other way I can highlight the changes?

annotation-utils/pom.xml
aakum03: There was a change to a url (that Liz helped me fix) that prevented maven from creating the build. However, the large diff should be due to reformatting.

annotation-utils/src/main/java/edu/isi/vista/annotationutils/AceCorpusTextSource.java
aakum03: Comments added

apf-ingester/apf_Ingester.py
aakum03: I actually need to delete this function, as it is not being used anymore. I will replace `str` with `Path` everywhere else

aakum03: Function removed (unecessary, as mentioned above)

annotation-utils/src/main/java/edu/isi/vista/annotationutils/RestoreJson.kt
```diff
@@ -1,109 +1,149 @@
...
+        }
+        fun restore(params: Parameters) {
+            val inputJsonDirectory = params.getExistingDirectory("inputJsonDirectory")!!
+            val outputDirectory = params.getCreatableDirectory("restoredJsonDirectory")!!
+
+            // Make objects for reading, parsing, and writing json:
+            val objectMapper = ObjectMapper()
+            val prettyPrinter = objectMapper.writerWithDefaultPrettyPrinter()
+            val gigaWordTextSource = makeTextSource(params, "gigaword")
+            val aceTextSource = makeTextSource(params, "ace")
```
aakum03, line 53: aceTextSource Added

annotation-utils/src/main/java/edu/isi/vista/annotationutils/RestoreJson.kt
```diff
@@ -1,109 +1,149 @@
...
+
+            inputJsonDirectory.walk().filter { it.isFile }.forEach { jsonFile ->
+                val filename = jsonFile.name
+                //returns docID if ace file, else returns null
```
aakum03, line 57: The following code block was changed to determine whether the document being processed from the stream is an ACE or gigaword document. The actual determination happens mostly in the `getAceDocID` function

annotation-utils/src/main/java/edu/isi/vista/annotationutils/RestoreJson.kt
```diff
@@ -1,109 +1,149 @@
...
+
+private fun Path.removePrefixPath(prefix: Path): Path {
+    require(this.startsWith(prefix)) {
+        "The path $this does not start with the prefix $prefix"
+    }
+    return this.subpath(prefix.nameCount, this.nameCount)
+}
+
+
+private fun getAceDocID(filename: String): String? {
```
aakum03, line 138: This function returns the event.subtype as a string if the document is an ACE document, else it returns null.
After some discussion with Liz, this regex seemed the a good option to differentiate ace and giga-word documents and extract the event detail at the same time. (Although currently, ace documents have `xmi`, in their name, using this to differentiate would not be future-proof)

PR #63, opened 2020-04-30
Update inception instructions

elizlee:


docs/SETTING_UP_INCEPTION_ON_ISI_VM.md
```diff
@@ -166,10 +166,45 @@ sudo systemctl restart inception
...
 # Update Inception
 
+Official documentation: https://inception-project.github.io/releases/0.15.2/docs/admin-guide.html#sect_upgrade
+
+## Create backups
+Before you begin the upgrade, you'll want to create backups in case the upgrade doesn't
+go as planned. Your backup files should consist of the Inception
+home directory and the SQL databases.
+
+1. Copy the home directory (`/srv/inception`) to `/nas/gaia/curated-training/inception_backups`.
```
elizlee, line 176: I might want to specify that the destination here is on `gaiadev01`

gabbard: That same path is mounted across our cluster, so no need.

PR #71, opened 2020-06-02
41 count sentences annotated

elizlee:
* Counts annotations by sentence instead of by document
* For ACE documents, counts sentences that contain clue words or additional triggers
* Includes # of event types annotated per project/language

### DIFF
--- a/annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExtractAnnotationStats.kt
+++ b/annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExtractAnnotationStats.kt
@@ -37,13 +37,7 @@ class ExtractAnnotationStats {
             val formatter = SimpleDateFormat("yyyy-MM-dd")
             val thisDate = formatter.format(currentDate.getTime())
 
-            // val documentsWithEventAnnotations: MutableList<DocumentAnnotation> = collectStats(exportAnnotationRoot)
             val sentencesWithEventAnnotations: MutableList<SentenceAnnotation> = collectStats(exportAnnotationRoot)
-
-            // From documentsWithEventAnnotations, get TOTAL, TOTAL PER USER and TOTAL PER EVENT TYPE
-//            val totalAnnotations = documentsWithEventAnnotations.size
-//            val annotationsByUser = documentsWithEventAnnotations.countBy { it.user }
-//            val annotationsByEventType = documentsWithEventAnnotations.countBy { it.eventType }
 
             // From sentencesWithEventAnnotations, got TOTAL, TOTAL PER USER, EVENT TYPE, AND CORPUS
             val positiveSentences = sentencesWithEventAnnotations.filter { !it.negativeExample }
@@ -53,7 +47,6 @@ class ExtractAnnotationStats {
             val annotationsByEventType = sentencesWithEventAnnotations.countBy { it.eventType }
             val positiveAnnotationsByCorpus = positiveSentences.countBy { it.corpus }
             val negativeAnnotationsByCorpus = negativeSentences.countBy { it.corpus }
-            logger.info("Negative annotations: $negativeAnnotationsByCorpus")
 
             // For better organization
             val sortedUsers = annotationsByUser.toSortedMap()
@@ -140,9 +133,6 @@ class ExtractAnnotationStats {
                                 val eventTypePattern = Regex(pattern = """\w+?-?\w+\.?\w+\.\w+""")
                                 eventType = eventTypePattern
                                         .find(input = folder)!!.value
-                                if (eventType.contains("SymptomOf")) {
-                                    logger.info("YAY!")
-                                }
                                 corpus = if (folder.contains("russian")) {
                                     "Russian"
                                 } else if (folder.contains("spanish")) {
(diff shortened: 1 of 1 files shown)


elizlee:
The new tables would look something like this:
Screen Shot 2020-06-08 at 10 49 10 AM

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExtractAnnotationStats.kt
```diff
@@ -290,7 +526,18 @@ class ExtractAnnotationStats {
...
+        val eventType: String,
+        val corpus: String,
+        val negativeExample: Boolean
+)
+data class AnnotationStats(
+        val total: Int,
+        val byUser: Map<String, Int?>,
+        val byEventType: Map<String, Int?>,
+        val byCorpusPositive: Map<String, Int?>?,
+        val byCorpusNegative: Map<String, Int?>?
```
berquist, line 541: Why are these all `Int?` and two of the maps `?` themselves? Can you make it so that if the values are going to be null, there just wouldn't be a key? And the mappings should never be null, they should be present but empty.

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExtractAnnotationStats.kt
```diff
@@ -227,11 +405,69 @@ class ExtractAnnotationStats {
...
+                                        td { +"${annotationDiffs.byCorpusNegative[corpus]}" }
+                                    }
+                                }
+                            }
+                        }
+                    }
+                    table {
+                        tr {
+                            th { +"Project" }
+                            th { +"Total event types" }
```
berquist, line 457: This is ambiguous to me, is it "total event types annotated all-time"?

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExtractAnnotationStats.kt
```diff
@@ -110,22 +136,150 @@ class ExtractAnnotationStats {
...
+        private fun getPrimarySpans(relations: JsonNode, spans: JsonNode): List<JsonNode> {
+            val dependents: MutableSet<String> = mutableSetOf()
+            val governors: MutableSet<String> = mutableSetOf()
+            val primarySpans: MutableSet<JsonNode> = mutableSetOf()
+            for (relation in relations) {
+                val relationDependent = relation["Dependent"].toString()
+                val relationGovernor = relation["Governor"].toString()
+                // A span is a primary trigger if it
+                // depends on nothing or itself.
+                // We prevent spans from being added to
```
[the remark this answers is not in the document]
elizlee: In our curated training relations, the governor is the argument

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExtractAnnotationStats.kt
elizlee: Yes; I will make these clearer

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExtractAnnotationStats.kt
```diff
@@ -34,16 +41,25 @@ class ExtractAnnotationStats {
             val formatter = SimpleDateFormat("yyyy-MM-dd")
             val thisDate = formatter.format(currentDate.getTime())
 
-            val documentsWithEventAnnotations: MutableList<DocumentAnnotation> = collectStats(exportAnnotationRoot)
+            val sentencesWithAnnotations: MutableList<SentenceAnnotation> = collectStats(exportAnnotationRoot)
+
+            // From sentencesWithEventAnnotations, got TOTAL, TOTAL PER USER, EVENT TYPE, AND CORPUS
+            val positiveSentences = sentencesWithAnnotations.filter { !it.negativeExample }
+            val negativeSentences = sentencesWithAnnotations.filter { it.negativeExample }
```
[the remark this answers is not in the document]
elizlee: Ah, thank you!

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExtractAnnotationStats.kt
```diff
@@ -110,22 +136,150 @@ class ExtractAnnotationStats {
...
+                // arguments are still considered primary triggers.
+                dependents.add(relationDependent)
+                if (relationDependent != relationGovernor) {
+                    governors.add(relationGovernor)
+                }
+            }
+            for (dependent in dependents) {
+                if (!governors.contains(dependent)) {
+                    primarySpans.add(spans[dependent])
+                }
```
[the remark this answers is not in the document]
elizlee: Adding a dependent to `primarySpans` if it's not in the set of governors in the first loop wouldn't necessarily filter out all secondary triggers.

For example, if there are two spans A and B where A is an argument of itself and a secondary trigger to B, combining these loops would incorrectly make A a primary trigger (Two relations: (dep=A, gov=A), (dep=B, gov=A)).

However, it *should* still yield the same sentence count since there is a check for duplicate sentences - it would just have another trigger to process.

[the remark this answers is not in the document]
berquist: This was hard for me to understand without writing out explicit pseudocode. Does @gabbard have an opinion on if there should be more of a comment here?

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExtractAnnotationStats.kt
```diff
@@ -161,17 +164,25 @@ class ExtractAnnotationStats {
...
+                                    if (documentRelations != null) {
+                                        significantDocumentSpans = getPrimarySpans(documentRelations, documentSpans)
+                                    }
+                                    // Add any argument-less spans from CTEventSpan.
+                                    // An example of a CTEventSpan value is
+                                    // "[1000, 1001, {"sofa": 1, "begin": 200, "end": 205, "negative_example": true}]"
+                                    // Integers represent spans in relations; we are
+                                    // only interested in collecting the spans here that aren't.
+                                    for (item in documentCTEventSpan) {
+                                        if (!item.isInt) {
```
berquist, line 179: Can the third item in `documentCTEventSpan` be used, instead of filtering on the type? Or will there be multiple spans in `documentCTEventSpan`?

elizlee: An argument-less span could be at any index, and there could be more than one.

elizlee: This might be a better example:
`[12135,12140,12145,{"sofa":1,"begin":310,"end":318,"negative_example":true},12130]`

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExtractAnnotationStats.kt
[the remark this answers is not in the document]
berquist: I think this is better, and if other people want to see it, they can write out all the code with explicit values like I did.

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExtractAnnotationStats.kt
berquist: I understand, and this also handles the case if there is ever more than one argument-less span. I brought it up because I was hoping there was a more concrete way to check the type of that bit of JSON, but if you can guarantee that `documentCTEventSpan` is only ever integers and that dict, I'm ok with this.

PR #74, opened 2020-06-15
73 collect times users spend on annotation

### LINKED ISSUE #73: Collect times users spend on annotation
We want to find out how much time each annotator has spent working on each project. Ideally we could calculate this data nightly.

It looks like the best source of time-tracking information is in each project's `event.log` file. An example showing activity on one document looks like this:
```
{"annotator":"laura_godlaski","created":1589578270000,"document_name":"CNNHL_ENG_20030416_133739.13-Business.Declare-Bankruptcy.xmi","event":"AfterCasWrittenEvent","id":717312,"user":"laura_godlaski"}
{"created":1589578270000,"details":{"state":"ANNOTATION_INPROGRESS","previousState":"NEW"},"document_name":"CNNHL_ENG_20030416_133739.13-Business.Declare-Bankruptcy.xmi","event":"DocumentStateChangedEvent","id":717313,"user":"laura_godlaski"}
{"annotator":"laura_godlaski","created":1589578270000,"details":{"state":"INPROGRESS","previousState":"NEW"},"document_name":"CNNHL_ENG_20030416_133739.13-Business.Declare-Bankruptcy.xmi","event":"AnnotationStateChangeEvent","id":717314,"user":"laura_godlaski"}
{"annotator":"laura_godlaski","created":1589578270000,"document_name":"CNNHL_ENG_20030416_133739.13-Business.Declare-Bankruptcy.xmi","event":"DocumentOpenedEvent","id":717315,"user":"laura_godlaski"}
{"annotator":"laura_godlaski","created":1589578287000,"document_name":"CNNHL_ENG_20030416_133739.13-Business.Declare-Bankruptcy.xmi","event":"AfterCasWrittenEvent","id":717316,"user":"laura_godlaski"}
{"annotator":"laura_godlaski","created":1589578287000,"document_name":"CNNHL_ENG_20030416_133739.13-Business.Declare-Bankruptcy.xmi","event":"DocumentOpenedEvent","id":717317,"user":"laura_godlaski"}
{"annotator":"laura_godlaski","created":1589578337000,"details":{"addr":4090,"begin":1320,"end":1328,"type":"webanno.custom.CTEventSpan","text":"file for"},"document_name":"CNNHL_ENG_20030416_133739.13-Business.Declare-Bankruptcy.xmi","event":"SpanCreatedEvent","id":717318,"user":"laura_godlaski"}
{"annotator":"laura_godlaski","created":1589578337000,"document_name":"CNNHL_ENG_20030416_133739.13-Business.Declare-Bankruptcy.xmi","event":"AfterCasWrittenEvent","id":717319,"user":"laura_godlaski"}
{"annotator":"laura_godlaski","created":1589578341000,"details":{"addr":8,"begin":1329,"end":1339,"type":"webanno.custom.CTEventSpan","text":"bankruptcy"},"document_name":"CNNHL_ENG_20030416_133739.13-Business.Declare-Bankruptcy.xmi","event":"RelationCreatedEvent","id":717320,"user":"laura_godlaski"}
{"annotator":"laura_godlaski","created":1589578341000,"document_name":"CNNHL_ENG_20030416_133739.13-Business.Declare-Bankruptcy.xmi","event":"AfterCasWrittenEvent","id":717321,"user":"laura_godlaski"}
{"annotator":"laura_godlaski","created":1589578345000,"document_name":"CNNHL_ENG_20030304_142751.10-Business.Declare-Bankruptcy.xmi","event":"AfterCasWrittenEvent","id":717322,"user":"laura_godlaski"}
```
We could estimate from the timestamps (the values in "created") in this set of events that Laura was working on this particular document for about 1 minute and 15 seconds.

Proposed plan:
* In `ExportAnnotations`, collect `event.log` from each project along with the annotations. I'll have to check whether the remote API allows this - otherwise I'll have to think of a different method...
* Write a new script - something like `GetUserAnnotationDurations` - that does the following:
  * For each user, for each project, read its`event.log`
  * Have a mapping of projects to documents to their total annotation times (e.g. `{"Conflict.Attack": {"APW_ENG_20030416.0135": 120}}`)
  * Run through each event - since the events are listed chronologically, get the "start" and "end" timestamps of a given document's activity and add the difference to the mapping.
    * Check the time passed between each event - if it passes a certain threshold (like 5-10 minutes?), it may suggest that the annotator took a break or stopped for the day, so it should not be added to the total difference.
  * For each user, create an output file that holds the total annotation durations in seconds and a human-readable format for each project, e.g.
```
{
  "Conflict.Attack": {
    "seconds": 23372,
    "formatted": "6h:29m:32s",
  },
  "Conflict.Demonstrate": {
    "seconds": 20727,
    "formatted": "5h:45m:27s",
  },
  ...
} 
```
* Should any time statistics be printed in the HTML email? If so, adding them with `ExtractAnnotationStats` would be another step.

elizlee:
Closes #73 

Main file: `GetAnnotationDurations.kt`

* Uses a project's `event.log` to estimate how long the user spent annotating on that project
* `ExportAnnotations` now saves each project's `event.log`
* `ExtractAnnotationStats` prints the resulting annotation times to stats reports
* Collecting the times will be part of the nightly export routine

### DIFF
--- a/annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExtractAnnotationStats.kt
+++ b/annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExtractAnnotationStats.kt
@@ -616,7 +616,7 @@ class ExtractAnnotationStats {
                                             td {}
                                         }
                                         td {+project}
-                                        td {+projectSentenceCount}
+                                        td {+"$projectSentenceCount"}
                                         td {+totalTime}
                                         if (secondsDiff != null){
                                             td {+secondsToHMS(secondsDiff)}
@@ -653,7 +653,6 @@ private fun getSentenceCountsByProject(sentenceList: List<SentenceAnnotation>): 
     val sentencesByUser = sentenceList.groupBy { it.user }
     val projectCounts = mutableMapOf<String, Int>()
     for (user in sentencesByUser) {
-        logger.info { "user = $user" }
         val username = user.key
         val eventTypeCountsForUser = user.value.countBy { it.eventType }
         for (eventType in eventTypeCountsForUser) {
--- a/annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExtractAnnotationStats.kt
+++ b/annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExtractAnnotationStats.kt
@@ -30,13 +30,13 @@ class ExtractAnnotationStats {
             if (argv.size != 1) {
                 throw RuntimeException("Expected a single argument: a parameter file")
             }
-            val paramsLoader = SerifStyleParameterFileLoader.Builder()
-                    .interpolateEnvironmentalVariables(true).build()
+            val paramsLoader = SerifStyleParameterFileLoader.Builder().build()
             val params = paramsLoader.load(File(argv[0]))
             extractStats(params)
         }
         fun extractStats(params: edu.isi.nlp.parameters.Parameters) {
             val exportAnnotationRoot = params.getExistingDirectory("exportedAnnotationRoot")
+            val timeReportRoot = params.getExistingDirectory("timeReportRoot")
             val statisticsDirectory = params.getExistingDirectory("statisticsDirectory")
             val currentDate = Calendar.getInstance(TimeZone.getTimeZone("UTC"))
             TimeZone.setDefault(TimeZone.getTimeZone("UTC"))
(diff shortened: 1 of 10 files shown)


elizlee:
The new table would look something like this:
Screen Shot 2020-06-17 at 10 44 10 AM

annotation-utils/src/main/java/edu/isi/vista/annotationutils/GetAnnotationDurations.kt
```diff
@@ -0,0 +1,217 @@
...
+                // This is the first event in the log
+                currentDocument = documentName
+            }
+            // If there is a relatively long gap between events,
+            // it may suggest that the annotator took a break.
+            // Check if the time elapsed since the last event
+            // is less than 5 minutes (300,000 milliseconds) -
+            // else, don't add it to the overall time spent
+            // on the document.
+            if (timeSinceLastEvent < 300000) {
```
elizlee, line 193: Does five minutes seem like an appropriate threshold? I want to account for breaks, but I also want to check for slow annotators. Considering this would be the time between individual events (marking a span, drawing relations, etc.) maybe this should be lower.

gabbard: I would lower it to 2 minutes.

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExtractAnnotationStats.kt
```diff
@@ -360,7 +405,13 @@ class ExtractAnnotationStats {
                         (corpus, newCount) ->
                         previousStats.byCorpusNegative.getValue(corpus).let { newCount.minus(it) }
                     }
-                return AnnotationStats(totalDiff, userDiff, eventTypeDiff, posCorpusDiff, negCorpusDiff)
+            val timesDiff = newStats.annotationTimes
```
gabbard, line 408: Can you add more comments about what is going on here?

PR #85, opened 2020-07-07
84 map search queries to annotation info

### LINKED ISSUE #84: ParseEventLogs: Map search queries to annotation info

elizlee:
Closes #84 

The indicator searches list now includes the spans marked during each indicator search. This data can be used to create indicator-to-sentence mappings.

### DIFF
--- a/annotation-utils/src/main/java/edu/isi/vista/annotationutils/ParseEventLogs.kt
+++ b/annotation-utils/src/main/java/edu/isi/vista/annotationutils/ParseEventLogs.kt
@@ -100,19 +100,34 @@ class ParseEventLogs {
                     logger.info { "Skipping $projectName"}
                 }
             }
-            // If there is a directory of original log files, include that data
+            // If there is a directory of original log files, include that data.
+            // First, create a mapping of eventType/usernames to project info
+            // in order to select the right project to modify.
+            val eventAndUserToProjectInfo = allProjectInfo.map {
+                Pair(it.eventType, it.username) to it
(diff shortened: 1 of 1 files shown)


annotation-utils/src/main/java/edu/isi/vista/annotationutils/ParseEventLogs.kt
```diff
@@ -100,24 +100,41 @@ class ParseEventLogs {
                     logger.info { "Skipping $projectName"}
                 }
             }
-            // If there is a directory of original log files, include that data
+            // If there is a directory of original log files, include that data.
+            // First, create a mapping of eventType/usernames to project info
+            // in order to select the right project to modify.
+            val eventAndUserToProjectInfo = allProjectInfo.map {
+                Pair(it.eventType, it.username) to it
+            }.toMap().toMutableMap()
```
joecummings, line 108: Can a `MutableMap` only be constructred from  `Map`?

joecummings (approved):
One question but not blocking 👍

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ParseEventLogs.kt
elizlee: Yeah. The `.map()` function here applies a lambda function and creates a `List` of `Pair` objects which can then be converted to a `Map` -- `toMutableMap()` can only be performed on a `Map` object.

PR #89, opened 2020-07-21
Add arguments to existing projects

### LINKED ISSUE #88: Add new argument tags to existing Inception projects
We would like to be able to add new sets of tags to existing projects without having to add them manually (which would be extremely tedious depending on how many projects must be modified). For example, an ontology extension gives all `Life.Die(.subtype)` event types a `Killer` argument role, so all applicable projects will need a `Killer` tag added if it doesn't have one already.

Since the Inception application doesn't appear to allow mass tag addition, I thought of this plan which uses the remote API:
1. **Export and save** all relevant project files (e.g. all `Life.Die` projects).
2. Create modified copies of these ZIP archives such that for each `.json` project file, **add the new tags** for that event type (e.g. add `Killer` to `tag_set[tags]`).
3. We want to import these modified projects without creating a bunch of `copy_of` projects, which is what Inception automatically does to prevent projects from being overwritten (and this doesn't appear to be configurable), so **delete** each project from Inception that was successfully modified locally.
4. For each modified project, **import** these to Inception.

elizlee:
Closes #88

### DIFF
--- a/annotation-utils/src/main/java/edu/isi/vista/annotationutils/AddNewTags.kt
+++ b/annotation-utils/src/main/java/edu/isi/vista/annotationutils/AddNewTags.kt
@@ -121,13 +121,14 @@ fun main(argv: Array<String>) {
     val projectsToEventTypes = mutableMapOf<Project, String>()
     for (serverProject in serverProjects) {
         val projectName = serverProject.name
-        var abridgedProjectName = serverProject.name
         // Ignore any language prefixes so that it knows
         // to modify *all* projects of this event type
-        for (language in languagePrefixes) {
-            if (projectName.startsWith(language)) {
-                abridgedProjectName = abridgedProjectName.removePrefix("$language-")
-            }
+        val abridgedProjectName = if (languagePrefixes.any {projectName.startsWith(it)}) {
+            val firstHyphen = projectName.indexOf('-')
+            val language = projectName.substring(0, firstHyphen)
+            projectName.removePrefix("$language-")
+        } else {
+            projectName
         }
         for (eventType in eventTypesToModify) {
             // If the project's name minus language prefix begins with the project name to edit:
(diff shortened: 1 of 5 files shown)


annotation-utils/src/main/java/edu/isi/vista/annotationutils/AddNewTags.kt
```diff
@@ -0,0 +1,284 @@
...
+            ?: throw java.lang.RuntimeException("Could not fetch projects from $inceptionUrl. Aborting")
+    logger.info { "Projects on server ${serverProjects.map { it.name }}" }
+    val importUrl = "$inceptionUrl/api/aero/v1/projects/import"
+    val failureSet = mutableSetOf<String>()  // will hold names of projects that were not modified
+
+    // Determine which projects need to be processed
+    val projectsToEventTypes = mutableMapOf<Project, String>()
+    for (serverProject in serverProjects) {
+        val projectName = serverProject.name
+        var abridgedProjectName = serverProject.name
```
joecummings, line 124: Maybe move this down into the conditional, so that is the only place that sets `abridgedProjectName`. 

```
If projectName.startsWith(language)) {
    abridged = projectName.removePrefix....
} else {
   abridged = projectName
```

annotation-utils/src/main/java/edu/isi/vista/annotationutils/AddNewTags.kt
```diff
@@ -0,0 +1,284 @@
...
+                retryOnFuelError {
+                    "$inceptionUrl/api/aero/v1/projects/${project.key.id}".httpDelete()
+                            .authenticateToInception()
+                            .retryOnResponseFailure()
+                }
+
+                // Step 4: import the new projects.
+                logger.info { "Importing project $modifiedZipOutputFile" }
+                retryOnFuelError {
+                    // The documentation is a lie.
```
joecummings, line 226: ?

elizlee: I don't think we need this 😅

PR #101, opened 2020-11-06
100 adjustments to accommodate username mapping

### LINKED ISSUE #100: Adjustments to accommodate username mapping
After analyzing the errors from the stats of the modified annotation data, the following adjustments need to be done:

1. Make sure that the appropriate annotator names are being replaced, as sometimes "admin" documents are found in another user's projects and end up overwriting the user's annotations on any similar documents.
2. Use the same username mapping to gather data in `ParseEventLogs`.

elizlee:
Closes #100 

Along with addressing two issues surrounding the username replacement, this fixes a block of code that prevented indicator searches from being saved.

### DIFF
--- a/annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExportAnnotations.kt
+++ b/annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExportAnnotations.kt
@@ -83,7 +83,7 @@ class ExportAnnotations {
 
             // load the usernames map if one is given
             val usernameMap = if (usernameJson != null)
-                mapper.readTree(usernameJson) as ObjectNode else null
+                mapper.readTree(usernameJson) as ObjectNode else mapper.readTree("{}") as ObjectNode
 
             // extension function to avoid authentication boilerplate
             fun Request.authenticateToInception() =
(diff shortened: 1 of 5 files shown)


annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExportAnnotations.kt
```diff
@@ -123,12 +123,9 @@ class ExportAnnotations {
...
                 var finalProjectName = project.name
-                var outputUsername = projectUsername
-                if (usernameMap != null) {
-                    outputUsername = usernameMap.get(projectUsername)?.toString()?.removeSurrounding("\"")
-                    if (projectUsername != null && outputUsername != null) {
-                        finalProjectName = project.name.substring(0, userSeparatorIndex + 1) + outputUsername
-                    }
+                val outputUsername = usernameMap.get(projectUsername)?.toString()?.removeSurrounding("\"")
+                if (projectUsername != null && outputUsername != null) {
+                    finalProjectName = project.name.substring(0, userSeparatorIndex + 1) + outputUsername
```
joecummings, line 128: Is there anywhere in here where the format of the files is explicitly said in a comment? Would be informative for this code

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ParseEventLogs.kt
```diff
@@ -229,7 +237,9 @@ private fun parseProjectEvents(logEvents: List<JsonNode>, username: String): Pai
     for (event in logEvents) {
         val inceptionEventType = event.get("event").toString().removeSurrounding("\"")
         val documentName = event.get("document_name")?.toString()?.removeSurrounding("\"")
-        val user = event.get("user")?.toString()?.removeSurrounding("\"")
+        val logUser = event.get("user")?.toString()?.removeSurrounding("\"")
+        val mappedUser = usernameMap.get(logUser)?.toString()?.removeSurrounding("\"")
+        val user = mappedUser ?: logUser
```
joecummings, line 242: Just curious - how does this `?:` work?

joecummings (approved):
One nit and one question but lgtm

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ExportAnnotations.kt
elizlee: I could certainly add it

annotation-utils/src/main/java/edu/isi/vista/annotationutils/ParseEventLogs.kt
elizlee: It's basically short for "if x is not null then x else y"
https://kotlinlang.org/docs/reference/null-safety.html#elvis-operator

elizlee: So here, if `mappedUser` is not null, `user` will be `mappedUser`, else it will be `logUser`.

PR #103, opened 2020-11-10
Skip exporting documents without annotations

### LINKED ISSUE #102: Filter out documents without annotations
Currently, any imported document that is opened by the annotator on Inception is considered "in progress", even if it contains no annotations; annotators may open documents for annotating and then realize that it is not markable. This leaves us with several documents being saved amongst our annotation data with nothing marked. To reduce clutter, we would like to apply a filter to `ExportAnnotations` so that it will skip saving a document if it has no marked spans.

elizlee:
Closes #102


joecummings (approved):
lgtm 👍

Issue #105, opened 2020-11-12
Changes to annotation data storage


elizlee:
To make sure our annotation repos remain under 1 GB, we want to make the following changes:
* Add an optional step to ExportAnnotations that compresses the exported JSON after the Inception export is complete.
* Add an optional step to the [Curated Training Ingester](https://github.com/isi-vista/gaia-event-extraction/blob/master/gaia_event_extraction/ingesters/curated_training_ingester.py) that compresses the converted FlexNLP docs.

Additionally, we want to change what is stored in each of the annotation repos:
* `curated-training-annotation`: the compressed exported JSON and the stats reports
* `curated-training-vistanlp`: the stats reports and the compressed FlexNLP documents (currently also includes the exported JSON)

elizlee:
Steps complete

PR #106, opened 2020-11-13
Add option to zip exports

elizlee:
This PR aims to support the changes to the annotation data storage described in #105 .

Connected to this is https://github.com/isi-vista/gaia-event-extraction/pull/213 .

Note that even if the user chooses the "compress" option, the uncompressed output will still be saved to the designated directory. I think it would make sense to eventually allow the other steps in the pipeline to be able to read from zip archives so that the output can be "either compressed or not," but I would rather handle this in another issue.


joecummings (approved):
lgtm

